PR #11, opened 2021-04-22
Build against OpenSearch 1.0.0-alpha1.

dblock:
Signed-off-by: dblock <dblock@amazon.com>

### Description
Build against OpenSearch 1.0.0-alpha1.

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,7 +26,7 @@
 
 buildscript {
     ext {
-        opensearch_version = System.getProperty("opensearch.version", "1.0.0")
+        opensearch_version = System.getProperty("opensearch.version", "1.0.0-alpha1")
     }
 
     repositories {
@@ -35,7 +35,7 @@ buildscript {
     }
 
     dependencies {
-        classpath "org.opensearch.gradle:build-tools:${opensearch_version}-SNAPSHOT"
+        classpath "org.opensearch.gradle:build-tools:${opensearch_version}"
     }
 }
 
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -87,7 +87,7 @@ testClusters.javaRestTest {
 
 allprojects {
     group = 'com.amazon.opendistroforelasticsearch'
-    version = "${opendistroVersion}.0"
+    version = "${opendistroVersion}"
```
saratvemulapalli, line 90: The last digit is intended for plugins to have their own minor version.
Could you help me understand why we'd like to remove it?

saratvemulapalli (commented):
Overall the changes look good to me.

Dropped a minor comment.

build.gradle
dblock: Hmm, it was generating an -alpha1.0 build. I can track where it's used incorrectly or we can leave it. WDYT?

saratvemulapalli: Yeah because `-alpha1` was appended to the `gradle.properties` file. I would prefer to stick with 1.0.0.0 as we have decided.

saratvemulapalli: Same here as common-utils, lets have the 4th digit for the minor version.

saratvemulapalli (approved):
Changes look good to me.

Just a minor comment.
Thanks a lot dB!

build.gradle
dblock: reverting

PR #13, opened 2021-04-23
Testing integration tests in CI

qreshi:

### Description

Use the `build` target that runs integration tests.
This PR is a duplicate of #12 but will be used for testing since the integ failures are happening on the GitHub Action workflow but not locally.
 
### Check List

### DIFF
--- a/.github/workflows/test-and-build-workflow.yml
+++ b/.github/workflows/test-and-build-workflow.yml
@@ -36,10 +36,10 @@ jobs:
         working-directory: ./OpenSearch
         run: ./gradlew publishToMavenLocal -Dbuild.version_qualifier=alpha1 -Dbuild.snapshot=false
         
-      # common-utils
+      # job-scheduler
       - name: Build and Test
         run: |
-          ./gradlew test -Dopensearch.version=1.0.0-alpha1 -Dbuild.snapshot=false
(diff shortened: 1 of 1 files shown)


dblock:
I'll close this since it is a dup of #12. Integ tests do fail locally if you build against the alpha1 tag of OpenSearch. Once we have an alpha2 tagged I'll update that PR and we'll get to a green state.

qreshi:
> I'll close this since it is a dup of #12. Integ tests do fail locally if you build against the alpha1 tag of OpenSearch. Once we have an alpha2 tagged I'll update that PR and we'll get to a green state.

I wasn't seeing consistent test results locally vs. on GitHub Actions even when using alpha1 tag for OpenSearch in both cases. It would *usually* pass locally but fail using GitHub Actions a lot.  I'm fine with checking again with alpha2 but it does feel like there could possibly be a bleed between tests that's more likely to occur when running with CI.

If we still see consistent failures with the CI workflow, we can try adding sleeps in the beginning of the tests to see if late cleanup operations are possibly interfering with the next test that runs. That's what I was planning to check with this duplicate PR but we can perform the same work in #12

dblock:
I added `@Ignore` on the tests and opened https://github.com/opensearch-project/job-scheduler/issues/15.

PR #14, opened 2021-04-24
Rename plugin

qreshi:

### Description

Updating plugin name to contain `opensearch`
 
### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,7 +45,7 @@ plugins {
 }
 
 ext {
-    opendistroVersion = "${version}"
+    opensearchVersion = "${version}"
     isSnapshot = "true" == System.getProperty("build.snapshot", "true")
 }
 
@@ -72,7 +72,7 @@ loggerUsageCheck.enabled = false
 
 opensearchplugin {
     name 'opensearch-job-scheduler'
-    description 'Open Distro for Elasticsearch job schduler plugin'
+    description 'OpenSearch Job Scheduler plugin'
     classname 'com.amazon.opendistroforelasticsearch.jobscheduler.JobSchedulerPlugin'
 }
 
@@ -87,7 +87,7 @@ testClusters.javaRestTest {
 
 allprojects {
     group = 'com.amazon.opendistroforelasticsearch'
-    version = "${opendistroVersion}.0-beta1"
+    version = "${opensearchVersion}.0-beta1"
 
     apply from: "$rootDir/build-tools/repositories.gradle"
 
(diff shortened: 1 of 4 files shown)


lezzago (approved):

build.gradle
```diff
@@ -87,7 +87,7 @@ testClusters.javaRestTest {
 
 allprojects {
     group = 'com.amazon.opendistroforelasticsearch'
-    version = "${opendistroVersion}.0"
+    version = "${opendistroVersion}.0-beta1"
```
dblock, line 90: Let's get rid of these `opendistroVersion` while we're at it?

build.gradle
```diff
@@ -71,7 +71,7 @@ validateNebulaPom.enabled = false
 loggerUsageCheck.enabled = false
 
 opensearchplugin {
-    name 'opendistro-job-scheduler'
+    name 'opensearch-job-scheduler'
     description 'Open Distro for Elasticsearch job schduler plugin'
```
dblock, line 75: Let's fix the description to be OpenSearch, including the `schduler` typo?

dblock (changes requested):

qreshi:
@dblock Addressed the comments.

I also had a couple questions.

1. What should we update this to in `build.gradle`?:
https://github.com/opensearch-project/job-scheduler/blob/0918a568d215d2b448cd21fcd350f20d38a1c7ed/build.gradle#L140-L146

2. Previously ODFE (OpenDistro for Elasticsearch) used to refer to the distribution whereas Elasticsearch referred to the backend search engine. Now, OpenSearch refers to both the distribution and the backend search engine. Do you think this could get confusing at all when referencing distribution versions/engine versions and the like?

dblock (approved):

dblock:
> @dblock Addressed the comments.
> 
> I also had a couple questions.
> 
> 1. What should we update this to in `build.gradle`?:
>    https://github.com/opensearch-project/job-scheduler/blob/0918a568d215d2b448cd21fcd350f20d38a1c7ed/build.gradle#L140-L146
> 2. Previously ODFE (OpenDistro for Elasticsearch) used to refer to the distribution whereas Elasticsearch referred to the backend search engine. Now, OpenSearch refers to both the distribution and the backend search engine. Do you think this could get confusing at all when referencing distribution versions/engine versions and the like?

I'd be guessing on both. I think we should standardize this soon,  /cc:@saratvemulapalli

saratvemulapalli:
> > @dblock Addressed the comments.
> > I also had a couple questions.
> > 
> > 1. What should we update this to in `build.gradle`?:
> >    https://github.com/opensearch-project/job-scheduler/blob/0918a568d215d2b448cd21fcd350f20d38a1c7ed/build.gradle#L140-L146
> > 2. Previously ODFE (OpenDistro for Elasticsearch) used to refer to the distribution whereas Elasticsearch referred to the backend search engine. Now, OpenSearch refers to both the distribution and the backend search engine. Do you think this could get confusing at all when referencing distribution versions/engine versions and the like?
> 
> I'd be guessing on both. I think we should standardize this soon, /cc:@saratvemulapalli

Yeah definitely a great question. There are multiple questions involved, which are open:

1. Url for downloads
2. Docs URL
3. Distinguishing between backend engine version and plugin version.

PR #20, opened 2021-05-06
Rename settings from OpenDistro and OpenSearch, with backwards compatibility, using fallback

dblock:
Signed-off-by: dblock <dblock@amazon.com>

### Description

1. Rename setting from opendistro to opensearch.
2. Use opendistro settings as fallback.
 
An alternative to #19, depends on https://github.com/opensearch-project/OpenSearch/pull/665.

### opensearch.yml Settings

Manually testing upgrade path from `config/opensearch.yml` settings:

Set `opendistro.jobscheduler.retry_count: 6` in `config./opensearch.yml`.

```
tar vfxz opensearch-1.0.0-beta1-linux-x64.tar.gz 
cd opensearch-1.0.0-beta1
./bin/opensearch-plugin install opensearch-job-scheduler-1.0.0.0-beta1.zip 
```

Run `./bin/opensearch`, it will show the value for this setting at 7.

### Node Settings (and upgrade from OpenDistro)

Install OpenDistro

```
wget https://d3g5vo6xdbdb9a.cloudfront.net/tarball/opendistro-elasticsearch/opendistroforelasticsearch-1.13.2-linux-x64.tar.gz
tar vfxz opendistroforelasticsearch-1.13.2-linux-x64.tar.gz
cd opendistroforelasticsearch-1.13.2
./bin/elasticsearch
```

Set `opendistro.jobscheduler.retry_count: 6`.

```
curl -X PUT "localhost:9200/_cluster/settings?pretty" -H 'Content-Type: application/json' -d'{"persistent":{"opendistro.jobscheduler.retry_count":6}}'
```

Stop OpenDistro, install OpenSearch on top (or copy data). I built just the minimal engine + job-scheduler from source with these changes.

```
tar vfxz opensearch-1.0.0-beta1-linux-x64.tar.gz 
cd opensearch-1.0.0-beta1
./bin/opensearch-plugin install opensearch-job-scheduler-1.0.0.0-beta1.zip 
rm -rf data
cp -R ../opendistroforelasticsearch-1.13.2/data .
```

At start the default values are used before settings are read.

```
[2021-05-06T19:55:17,649][INFO ][c.a.o.j.s.JobSweeper     ] [ip-172-31-41-248] Background sweep search backoff retry count: 3
```

I thought this was a bug, but both old and new are just defaulted before they are read from the node configuration when it joins the cluster (with some debugging code).

```
[2021-05-06T19:55:17,651][INFO ][c.a.o.j.s.JobSweeper     ] [ip-172-31-41-248] JobSchedulerSettings.SWEEP_BACKOFF_RETRY_COUNT: 3
[2021-05-06T19:55:17,651][INFO ][c.a.o.j.s.JobSweeper     ] [ip-172-31-41-248] LegacyOpenDistroJobSchedulerSettings.SWEEP_BACKOFF_RETRY_COUNT: 3
```

As the node joins the cluster, it updates the value from the node configuration.

```
[2021-05-06T19:55:18,696][INFO ][o.o.c.s.ClusterSettings  ] [ip-172-31-41-248] updating [opensearch.jobscheduler.retry_count] from [3] to [6]
```

### Check List

  - [ ] All tests pass
- [ ] Passing build, needs changes in OpenSearch 

### DIFF
--- a/.github/workflows/test-and-build-workflow.yml
+++ b/.github/workflows/test-and-build-workflow.yml
@@ -30,20 +30,20 @@ jobs:
         uses: actions/checkout@v2
         with:
           repository: 'opensearch-project/OpenSearch'
-          ref: '1.0.0-beta1'
+          ref: '1.x'
           path: OpenSearch
       - name: Build OpenSearch
         working-directory: ./OpenSearch
-        run: ./gradlew publishToMavenLocal -Dbuild.version_qualifier=beta1 -Dbuild.snapshot=false
+        run: ./gradlew publishToMavenLocal -Dbuild.snapshot=false
         
       # job-scheduler
       - name: Build and Test
         run: |
-          ./gradlew build -Dopensearch.version=1.0.0-beta1 -Dbuild.snapshot=false
+          ./gradlew build -Dopensearch.version=1.0.0 -Dbuild.snapshot=false
 
       - name: Publish to Maven Local
         run: |
-          ./gradlew publishToMavenLocal -Dopensearch.version=1.0.0-beta1 -Dbuild.snapshot=false
+          ./gradlew publishToMavenLocal -Dopensearch.version=1.0.0 -Dbuild.snapshot=false
           
       - name: Upload Coverage Report
         uses: codecov/codecov-action@v1
--- a/src/main/java/com/amazon/opendistroforelasticsearch/jobscheduler/sweeper/JobSweeper.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/jobscheduler/sweeper/JobSweeper.java
@@ -136,11 +136,17 @@ public class JobSweeper extends LifecycleListener implements IndexingOperationLi
 
     private void loadSettings(Settings settings) {
         this.sweepPeriod = JobSchedulerSettings.SWEEP_PERIOD.get(settings);
+        log.info("Background full sweep with period: " + this.sweepPeriod.getMinutes());
         this.sweepPageMaxSize = JobSchedulerSettings.SWEEP_PAGE_SIZE.get(settings);
+        log.info("Background sweep page size: " + this.sweepPageMaxSize);
         this.sweepSearchTimeout = JobSchedulerSettings.REQUEST_TIMEOUT.get(settings);
+        log.info("Background sweep search timeout: " + this.sweepSearchTimeout.getMinutes());
         this.sweepSearchBackoffMillis = JobSchedulerSettings.SWEEP_BACKOFF_MILLIS.get(settings);
+        log.info("Background sweep search backoff: " + this.sweepSearchBackoffMillis.getMillis());
         this.sweepSearchBackoffRetryCount = JobSchedulerSettings.SWEEP_BACKOFF_RETRY_COUNT.get(settings);
+        log.info("Background sweep search backoff retry count: " + this.sweepSearchBackoffRetryCount);
         this.jitterLimit = JobSchedulerSettings.JITTER_LIMIT.get(settings);
+        log.info("Background sweep jitter limit: " + this.jitterLimit);
         this.sweepSearchBackoff = this.updateRetryPolicy();
     }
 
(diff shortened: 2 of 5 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/jobscheduler/JobSchedulerSettingsTests.java
```diff
@@ -0,0 +1,119 @@
...
+import org.junit.Before;
+import org.opensearch.common.settings.Setting;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.common.unit.TimeValue;
+import org.opensearch.test.OpenSearchTestCase;
+
+@SuppressWarnings({"rawtypes"})
+public class JobSchedulerSettingsTests extends OpenSearchTestCase {
+
+    JobSchedulerPlugin plugin = null;
```
Bukhtawar, line 41: nit: we can skip the null assignment

src/main/java/com/amazon/opendistroforelasticsearch/jobscheduler/sweeper/JobSweeper.java
```diff
@@ -148,25 +148,36 @@ private void addConfigListeners() {
...
                     sweepPeriod = timeValue;
-                    log.debug("Reinitializing background full sweep with period: " + sweepPeriod.getMinutes());
+                    log.debug("Reinitializing background full sweep with period: " + this.sweepPeriod.getMinutes());
                     initBackgroundSweep();
                 });
         clusterService.getClusterSettings().addSettingsUpdateConsumer(JobSchedulerSettings.SWEEP_PAGE_SIZE,
-                intValue -> sweepPageMaxSize = intValue);
+                intValue -> {
+                    sweepPageMaxSize = intValue;
+                    log.debug("Setting background sweep page size: " + this.sweepPageMaxSize);
```
Bukhtawar, line 157:
```suggestion
                    log.debug("Setting background sweep page size: {}" , this.sweepPageMaxSize);
```

src/test/java/com/amazon/opendistroforelasticsearch/jobscheduler/JobSchedulerSettingsTests.java
```diff
@@ -0,0 +1,119 @@
...
+
+    public void testSettingsGetValueWithLegacyFallback() {
+        Settings settings = Settings.builder()
+            .put("opendistro.jobscheduler.request_timeout", "1s")
+            .put("opendistro.jobscheduler.sweeper.backoff_millis", "2ms")
+            .put("opendistro.jobscheduler.retry_count", 3)
+            .put("opendistro.jobscheduler.sweeper.period", "4s")
+            .put("opendistro.jobscheduler.sweeper.page_size", 5)
+            .put("opendistro.jobscheduler.jitter_limit", 6)
+        .build();
```
Bukhtawar, line 101: Lets use LegacyOpenDistroJobSchedulerSettings constants here maybe

Bukhtawar (commented):
Thanks @dblock for the changes, Overall LGTM
1. Would it make sense to add ITs to this and see how would settings update via API behave with fallback. 
2. While this is good was wondering if we simply had a way to say `key.replaceFirst("^opendisto", "opensearch") `

dblock:
> Thanks @dblock for the changes, Overall LGTM
> 
> 1. Would it make sense to add ITs to this and see how would settings update via API behave with fallback.

Probably. I'll take a look at what's possible.

> 2. While this is good was wondering if we simply had a way to say `key.replaceFirst("^opendisto", "opensearch") `

The settings infrastructure supports upgrades, fallbacks, archiving settings, and a lot more. So some kind of `s/opendistro/opensearch/g` just felt too magical to me, and smelled like something that would break in very unexpected ways. It also wouldn't easily provide a visible deprecation path, such as existing well-understood built-in warnings or mechanisms designed and tested over many versions.

.github/workflows/test-and-build-workflow.yml
```diff
@@ -30,20 +30,20 @@ jobs:
         uses: actions/checkout@v2
         with:
           repository: 'opensearch-project/OpenSearch'
-          ref: '1.0.0-beta1'
+          ref: '1.x'
           path: OpenSearch
       - name: Build OpenSearch
         working-directory: ./OpenSearch
-        run: ./gradlew publishToMavenLocal -Dbuild.version_qualifier=beta1 -Dbuild.snapshot=false
```
saratvemulapalli, line 37: From the changes, I see we are leveraging `fallbackSetting` to take care of legacy settings. 
Ref: https://github.com/opensearch-project/OpenSearch/blob/0ba0e7cc26060f964fcbf6ee45bae53b3a9941d0/server/src/main/java/org/opensearch/common/settings/Setting.java#L1703:38

Moving to 1.x might not be needed. 
But eventually we will have to build out of 1.x. 

Let me know your thoughts on this, I propose:
1. cut a branch for `1.0.0.0-beta1` for all the plugins.
2. `1.0.0.0-beta1` will still build out of OpenSearch `1.0.0-beta1`, this helps having stable builds.
3. `main` will build out of the latest feature branch on OpenSearch, in this case `1.x`, this helps absorbing changes from dependencies sooner rather than later.

saratvemulapalli (approved):
Overall the changes look good the me!

Thanks for the change!

src/test/java/com/amazon/opendistroforelasticsearch/jobscheduler/JobSchedulerSettingsTests.java
dblock: I actually would rather not. If I were to make a mistake or a typo in the actual rename in the setting, or someone accidentally bulk replaces opendistro by openseaerch, the test will point to a behavior change or start failing. Which is what we really want. WDYT?

dblock:
@Bukhtawar I tried to write an integration test across multiple versions of ODFE/ES/OpenSearch, but that looks complicated. I opened https://github.com/opensearch-project/OpenSearch/issues/670, if you know how to best do that pls add comments.

PR #61, opened 2021-09-15
Adds a delay parameter to the job scheduler

downsrob:
Signed-off-by: Clay Downs downsrob@amazon.com

### Description
Adds a delay parameter to the job scheduler which allows plugins to put off execution time by a fixed number of milliseconds. For example, an interval schedule which starts at 10 AM with a delay of 3 minutes and an interval of 7 minutes would run at 10:03 AM, then 10:10 AM, then 10:17 AM and so on. This parameter is being added to enable plugins to easily add a delay to job schedules without parsing and modifying the user defined interval or cron schedule. 

This [index state management PR](https://github.com/opensearch-project/index-management/pull/147) depends on this job scheduler change to add the delay feature to ISM rollup jobs.

### Check List
  - [X] All tests pass

### DIFF
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -49,7 +49,7 @@ Then you will find the built artifact located at `build/distributions` directory
 ## Install
 Once you have built the plugin from source code, run
 ```bash
-opensearch-plugin install file://build/distributions/opensearch-job-scheduler-1.2.0.0-SNAPSHOT.zip
+opensearch-plugin install file://${PLUGIN_ZIP_FILE_PATH}
 ```
 to install the JobScheduler plugin to your OpenSearch.
 
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -49,7 +49,7 @@ Then you will find the built artifact located at `build/distributions` directory
 ## Install
 Once you have built the plugin from source code, run
 ```bash
-opensearch-plugin install file://PLUGIN_ZIP_FILE_PATH
+opensearch-plugin install file://build/distributions/opensearch-job-scheduler-1.2.0.0-SNAPSHOT.zip
 ```
 to install the JobScheduler plugin to your OpenSearch.
 
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -49,7 +49,7 @@ Then you will find the built artifact located at `build/distributions` directory
 ## Install
 Once you have built the plugin from source code, run
 ```bash
-opensearch-plugin install file://${PLUGIN_ZIP_FILE_PATH}
+opensearch-plugin install file://PLUGIN_ZIP_FILE_PATH
 ```
 to install the JobScheduler plugin to your OpenSearch.
 
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -49,7 +49,7 @@ Then you will find the built artifact located at `build/distributions` directory
 ## Install
 Once you have built the plugin from source code, run
 ```bash
-opensearch-plugin install file://${PLUGIN_ZIP_FILE_PATH}
+opensearch-plugin install file://build/distributions/opensearch-job-scheduler-1.1.0.0-SNAPSHOT.zip
 ```
 to install the JobScheduler plugin to your OpenSearch.
 
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -49,7 +49,7 @@ Then you will find the built artifact located at `build/distributions` directory
 ## Install
 Once you have built the plugin from source code, run
 ```bash
-opensearch-plugin install file://build/distributions/opensearch-job-scheduler-1.1.0.0-SNAPSHOT.zip
+opensearch-plugin install file:///path/to/target/releases/opensearch-job-scheduler-<version>.zip
 ```
 to install the JobScheduler plugin to your OpenSearch.
 
--- a/spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/CronSchedule.java
+++ b/spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/CronSchedule.java
@@ -103,6 +103,10 @@ public class CronSchedule implements Schedule {
         return this.expression;
     }
 
+    public Long getDelay() { return this.scheduleDelay; }
+
+    public void setDelay(Long delay) { this.scheduleDelay = delay; }
+
     @Override
     public Instant getNextExecutionTime(Instant time) {
         Instant baseTime = time == null ? this.clock.instant() : time;
--- a/spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/CronSchedule.java
+++ b/spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/CronSchedule.java
@@ -103,16 +103,16 @@ public class CronSchedule implements Schedule {
         return this.expression;
     }
 
-    public long getDelay() { return this.scheduleDelay == null ? 0 : this.scheduleDelay; }
-
-    public void setDelay(long delay) { this.scheduleDelay = delay; }
-
     @Override
     public Instant getNextExecutionTime(Instant time) {
         Instant baseTime = time == null ? this.clock.instant() : time;
         long delay = scheduleDelay == null ? 0 : scheduleDelay;
-
-        // the executionTime object doesn't use the delay, need to remove the delay before and then add it back after
+        // The executionTime object doesn't know about the delay, so first subtract the delay from the baseTime in case
+        // this moves to the previous interval, then add the delay to the returned execution time to get the correct time.
+        // For example, say it is 10:07 AM with an hourly schedule and a delay of 15 minutes. The next execution time
+        // should be 10:15 AM, but executionTime.nextExecution( 10:07 AM ) would return the next execution as 11 AM.
+        // By subtracting the delay first, the ExecutionTime object is given the input time as 9:52 AM, it returns
+        // 10:00 AM, and after adding the delay, we get the correct next execution time of 10:15 AM.
         ZonedDateTime zonedDateTime = ZonedDateTime.ofInstant(baseTime.minusMillis(delay), this.timezone);
         ZonedDateTime nextExecutionTime = this.executionTime.nextExecution(zonedDateTime).orElse(null);
 
--- a/spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/CronSchedule.java
+++ b/spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/CronSchedule.java
@@ -61,17 +61,26 @@ public class CronSchedule implements Schedule {
     private String expression;
     private ExecutionTime executionTime;
     private Clock clock;
+    private long scheduleDelay;
 
     public CronSchedule(String expression, ZoneId timezone) {
         this.expression = expression;
         this.timezone = timezone;
         this.executionTime = ExecutionTime.forCron(cronParser.parse(this.expression));
         clock = Clock.system(timezone);
+        this.scheduleDelay = 0;
+    }
+
+    public CronSchedule(String expression, ZoneId timezone, long scheduleDelay) {
+        this(expression, timezone);
+        this.scheduleDelay = scheduleDelay;
     }
 
     public CronSchedule(StreamInput input) throws IOException {
         timezone = input.readZoneId();
         expression = input.readString();
+        Long delayIn = input.readOptionalLong();
+        scheduleDelay = delayIn == null ? 0 : delayIn;
         executionTime = ExecutionTime.forCron(cronParser.parse(expression));
         clock = Clock.system(timezone);
     }
@@ -96,19 +105,24 @@ public class CronSchedule implements Schedule {
         return this.expression;
     }
 
+    @VisibleForTesting
+    public long getDelay() {
+        return this.scheduleDelay;
+    }
+
     @Override
     public Instant getNextExecutionTime(Instant time) {
         Instant baseTime = time == null ? this.clock.instant() : time;
 
-        ZonedDateTime zonedDateTime = ZonedDateTime.ofInstant(baseTime, this.timezone);
+        ZonedDateTime zonedDateTime = ZonedDateTime.ofInstant(baseTime.minusMillis(this.scheduleDelay), this.timezone);
         ZonedDateTime nextExecutionTime = this.executionTime.nextExecution(zonedDateTime).orElse(null);
 
-        return nextExecutionTime == null ? null : nextExecutionTime.toInstant();
+        return nextExecutionTime == null ? null : nextExecutionTime.toInstant().plusMillis(this.scheduleDelay);
     }
 
     @Override
     public Duration nextTimeToExecute() {
-        Instant now = this.clock.instant();
+        Instant now = this.clock.instant().minusMillis(this.scheduleDelay);
         ZonedDateTime zonedDateTime = ZonedDateTime.ofInstant(now, this.timezone);
         Optional<Duration> timeToNextExecution = this.executionTime.timeToNextExecution(zonedDateTime);
         return timeToNextExecution.orElse(null);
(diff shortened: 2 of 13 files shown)


spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/CronSchedule.java
```diff
@@ -172,17 +187,19 @@ public boolean equals(Object o) {
...
     public int hashCode() {
-        return Objects.hash(timezone, expression);
+        return Objects.hash(timezone, expression, scheduleDelay);
     }
 
     @Override
     public void writeTo(StreamOutput out) throws IOException {
         out.writeZoneId(timezone);
         out.writeString(expression);
+        out.writeOptionalLong(scheduleDelay);
```
thalurur, line 203: does this need to be optional? since the class is using the primitive type for the `scheduleDelay` this field should always be present

spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/CronSchedule.java
```diff
@@ -61,17 +61,26 @@
...
+    public CronSchedule(String expression, ZoneId timezone, long scheduleDelay) {
+        this(expression, timezone);
+        this.scheduleDelay = scheduleDelay;
     }
 
     public CronSchedule(StreamInput input) throws IOException {
         timezone = input.readZoneId();
         expression = input.readString();
+        Long delayIn = input.readOptionalLong();
+        scheduleDelay = delayIn == null ? 0 : delayIn;
```
thalurur, line 83: same comment as in the `writeTo` if we write to stream as non optional we don't need to do this, and since this is always present we can write as non optional in `writeTo` and read it as non optional.

spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/CronSchedule.java
```diff
@@ -61,17 +61,26 @@
...
         this.expression = expression;
         this.timezone = timezone;
         this.executionTime = ExecutionTime.forCron(cronParser.parse(this.expression));
         clock = Clock.system(timezone);
+        this.scheduleDelay = 0;
+    }
+
+    public CronSchedule(String expression, ZoneId timezone, long scheduleDelay) {
+        this(expression, timezone);
+        this.scheduleDelay = scheduleDelay;
```
thalurur, line 76: minor/nitpick:

Can you move the setting logic to all fields to new method `CronSchedule(String expression, ZoneId timezone, long scheduleDelay)` and call this method with scheduleDelay as zero from `CronSchedule(String expression, ZoneId timezone)` that way we are not setting the schedule delay twice when method at line 74 is called.

spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/CronSchedule.java
```diff
@@ -96,19 +105,24 @@ String getCronExpression() {
         return this.expression;
     }
 
+    public long getDelay() { return this.scheduleDelay; }
+
+    public void setDelay(long delay) { this.scheduleDelay = delay; }
```
thalurur, line 110: curious: why is the set method needed? we don't change the delay once its set during initialization right?

spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/CronSchedule.java
```diff
@@ -96,19 +105,24 @@ String getCronExpression() {
...
+
     @Override
     public Instant getNextExecutionTime(Instant time) {
         Instant baseTime = time == null ? this.clock.instant() : time;
 
-        ZonedDateTime zonedDateTime = ZonedDateTime.ofInstant(baseTime, this.timezone);
+        // the executionTime object doesn't use the delay, need to remove the delay before and then add it back after
+        ZonedDateTime zonedDateTime = ZonedDateTime.ofInstant(baseTime.minusMillis(this.scheduleDelay), this.timezone);
```
thalurur, line 117: I am not sure I follow this - why do we need to remove from baseTime and add to executionTime later?

spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/IntervalSchedule.java
```diff
@@ -66,36 +66,49 @@
...
+        this.scheduleDelay = scheduleDelay;
     }
 
     public IntervalSchedule(StreamInput input) throws IOException {
-        startTime = input.readInstant();
+        initialStartTime = input.readInstant();
         interval = input.readInt();
         unit = input.readEnum(ChronoUnit.class);
+        Long delayIn = input.readOptionalLong();
+        scheduleDelay = delayIn == null ? 0 : delayIn;
```
thalurur, line 103: same comment as `CronSchedule`

spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/IntervalSchedule.java
```diff
@@ -66,36 +66,49 @@
...
         this.intervalInMillis = Duration.of(interval, this.unit).toMillis();
         this.clock = Clock.system(ZoneId.systemDefault());
+        this.scheduleDelay = 0;
+    }
+
+    public IntervalSchedule(Instant startTime, int interval, ChronoUnit unit, long scheduleDelay) {
+        this(startTime, interval, unit);
+        this.startTimeWithDelay = startTime.plusMillis(scheduleDelay);
+        this.scheduleDelay = scheduleDelay;
     }
```
thalurur, line 96: same comment as `CronSchedule`

spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/CronSchedule.java
downsrob: When initializing a rollup the delay parameter is included separately from the schedule, so I use this setter to add the delay into the schedule after it is created by the parser. The use of the setter is here: https://github.com/opensearch-project/index-management/pull/147/files#diff-8aaa165b68190cf1eaab9b0c167746357d53f7b23b2ffef9046e5ba51c2cab23
Do you think that it would be better to just create a new schedule with the delay instead of modifying the delay of an existing schedule?

spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/CronSchedule.java
downsrob: I did it this way to isolate the delay to the CronSchedule class rather than modifying the ExecutionTime object to include the delay parameter. 
For example, say it is 10:07 AM and I have a cron schedule set up to run every hour with a delay of 15 minutes. Given an input time of 10:07 AM, the ExecutionTime object would return the next execution as 11 AM. By subtracting the delay first, the ExecutionTime object is given the input time as 9:52 AM, and it returns the next execution time of 10:00 AM. After adding the delay, we get the correct next execution time of 10:15 AM.
It is probably too confusing though, do you think it would be better to just include the delay in the ExecutionTime class to avoid all of this jumping around?

spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/CronSchedule.java
downsrob: This definitely didn't need to be optional with the primitive scheduleDelay, but I have since changed scheduleDelay to be a Long object as part of making the job scheduler backwards compatible. Now, if scheduleDelay is null then it isn't included in the streams or jsons, so it shouldn't blow up the other plugins which use the job scheduler without a delay. The downside is that in a plugin using delay, you would always need to use the constructor with the delay, even if you are just setting it to 0, as otherwise the streams and jsons wouldn't include the delay value.

thalurur: > so it shouldn't blow up the other plugins which use the job scheduler without a delay
What do you mean by this - how will this affect other plugins? the logic of constructing and deconstructing the objects is limited to just these classes not sure I am seeing what is the concern with using `long` and including it in bytestream always?

spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/Schedule.java
```diff
@@ -65,4 +66,18 @@
...
+     * @return the delay parameter of the schedule as a Long.
+     */
+    Long getDelay();
+
+    /**
+     * Sets the delay parameter of the schedule.
+     *
+     * @param delay the delay parameter.
+     */
+    void setDelay(Long delay);
```
zengyan-amazon, line 82: do we want to allow plugins to dynamically change the delay after initialization? that may cause the schedule harder to track

DEVELOPER_GUIDE.md
```diff
@@ -49,7 +49,7 @@ Then you will find the built artifact located at `build/distributions` directory
 ## Install
 Once you have built the plugin from source code, run
 ```bash
-opensearch-plugin install file://${PLUGIN_ZIP_FILE_PATH}
+opensearch-plugin install file://build/distributions/opensearch-job-scheduler-1.1.0.0-SNAPSHOT.zip
```
thalurur, line 52: I believe this is incorrect, it should be reverted. If this is failing link checker we might need to skip this file checked(?)

spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/CronSchedule.java
```diff
@@ -86,59 +93,71 @@ void setExecutionTime(ExecutionTime executionTime) {
...
-    @VisibleForTesting
-    String getCronExpression() {
+    public String getCronExpression() {
         return this.expression;
     }
 
+    public Long getDelay() { return this.scheduleDelay; }
+
+    @VisibleForTesting
+    public void setDelay(Long delay) { this.scheduleDelay = delay; }
```
thalurur, line 107: remove public?

DEVELOPER_GUIDE.md
thalurur: it looks like all the plugins removed install section from the DEVELOPER_GUIDE. 

I see one example in security plugin which tries to specify the path, you can try this to see if setting file path to something like this prevent link checker failure - https://github.com/opensearch-project/security-dashboards-plugin/blob/main/DEVELOPER_GUIDE.md#install

spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/CronSchedule.java
downsrob: I removed the setter, but currently the public getter is still needed in [rollup.kt](https://github.com/opensearch-project/index-management/pull/147/files#diff-8aaa165b68190cf1eaab9b0c167746357d53f7b23b2ffef9046e5ba51c2cab23) as a method to check if the job schedule delay matches the rollup job delay

thalurur: I was referring to just the setter

thalurur (dismissed):

spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/IntervalSchedule.java
```diff
@@ -178,21 +197,27 @@ public boolean equals(Object o) {
...
 
     @Override
     public int hashCode() {
-        return Objects.hash(startTime, interval, unit, intervalInMillis);
+        if (scheduleDelay == null) {
+            return Objects.hash(initialStartTime, interval, unit, intervalInMillis);
+        } else {
+            return Objects.hash(initialStartTime, interval, unit, intervalInMillis, scheduleDelay);
+        }
```
aditjind, line 213: I like use of Ternary operators! Would recommend using them here, as well.

spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/IntervalSchedule.java
```diff
@@ -107,22 +116,31 @@ public ChronoUnit getUnit() {
...
         Instant currentTime = this.clock.instant();
         long delta = currentTime.toEpochMilli() - enabledTimeEpochMillis;
-        long remainingScheduleTime = intervalInMillis - (delta % intervalInMillis);
-        return Duration.of(remainingScheduleTime, ChronoUnit.MILLIS);
+        if (delta >= 0) {
+            long remainingScheduleTime = intervalInMillis - (delta % intervalInMillis);
+            return Duration.of(remainingScheduleTime, ChronoUnit.MILLIS);
+        } else {
+            return Duration.ofMillis(enabledTimeEpochMillis - currentTime.toEpochMilli());
+        }
```
aditjind, line 143: Same comment on Ternary operators! Simplification of code increases its readability.

DEVELOPER_GUIDE.md
```diff
@@ -49,7 +49,7 @@ Then you will find the built artifact located at `build/distributions` directory
 ## Install
 Once you have built the plugin from source code, run
 ```bash
-opensearch-plugin install file://${PLUGIN_ZIP_FILE_PATH}
+opensearch-plugin install file:///path/to/target/releases/opensearch-job-scheduler-<version>.zip
```
aditjind, line 52: What are your thoughts on adding the example path to target?

spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/CronSchedule.java
```diff
@@ -147,16 +163,17 @@ public Boolean runningOnTime(Instant lastExecutionTime) {
...
 
     @Override
     public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
         builder.startObject()
                 .startObject(CRON_FIELD)
                 .field(EXPRESSION_FIELD, this.expression)
-                .field(TIMEZONE_FIELD, this.timezone.getId())
-                .endObject()
+                .field(TIMEZONE_FIELD, this.timezone.getId());
+        if (this.scheduleDelay != null) { builder.field(DELAY_FIELD, this.scheduleDelay); }
```
aditjind, line 175: Format this code maybe?

spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/CronSchedule.java
```diff
@@ -172,17 +189,23 @@ public boolean equals(Object o) {
...
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(timezone, expression);
+        if (scheduleDelay == null) {
+            return Objects.hash(timezone, expression);
+        } else {
+            return Objects.hash(timezone, expression, scheduleDelay);
+        }
```
aditjind, line 202: (Rolls eyes)! Ternary operators.

spi/src/test/java/org/opensearch/jobscheduler/spi/schedule/IntervalScheduleTests.java
```diff
@@ -146,17 +201,28 @@ public void testToXContent() throws IOException {
...
         IntervalSchedule intervalScheduleThree = new IntervalSchedule(Instant.ofEpochMilli(epochMilli), 4, ChronoUnit.MINUTES);
+        IntervalSchedule intervalScheduleFour = new IntervalSchedule(Instant.ofEpochMilli(epochMilli), 4, ChronoUnit.MINUTES, 5000);
+        IntervalSchedule intervalScheduleFive = new IntervalSchedule(Instant.ofEpochMilli(epochMilli), 4, ChronoUnit.MINUTES, 5000);
 
         Assert.assertEquals(intervalScheduleOne, intervalScheduleTwo);
         Assert.assertNotEquals(intervalScheduleOne, intervalScheduleThree);
         Assert.assertEquals(intervalScheduleOne.hashCode(), intervalScheduleTwo.hashCode());
+        Assert.assertNotEquals(intervalScheduleOne, intervalScheduleFour);
+        Assert.assertEquals(intervalScheduleFour, intervalScheduleFive);
+        Assert.assertEquals(intervalScheduleFour.hashCode(), intervalScheduleFive.hashCode());
```
aditjind, line 225: What are your thoughts on error messages in these assertions?

aditjind:
I have taken a cursory pass! I have minor comments. If this is a blocker for a release, I can approve it. O/w we can address these comments and get this one into production then.

spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/Schedule.java
aditjind: I am curious, do we need to track the schedule of individual jobs for each of the plugins? I can foresee some contention is cases where one plugin needs to schedule some jobs and another plugin which depends on the same job.

DEVELOPER_GUIDE.md
downsrob: I don't like this very much, but this follows another plugin: https://github.com/opensearch-project/security-dashboards-plugin/blob/ad6e723d4710b91cbcf1f0ec543b6ec24b976c53/DEVELOPER_GUIDE.md and this change lets the lychee link checker succeed

spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/IntervalSchedule.java
downsrob: I am also a big fan of ternary operators! But I try to only use them in simple cases, and I really try to not include anything that takes much thought. I agree with you on the other cases but I will leave this one as an if/else for that reason

thalurur (approved):

aditjind (approved):

PR #84, opened 2021-11-05
Add release notes for 1.2.0.0

qreshi:

### Description
Add release notes for 1.2.0.0 release.
 
### Issues Resolved
[List any issues this PR will resolve]
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/release-notes/opensearch.job-scheduler.release-notes-1.2.0.0.md
+++ b/release-notes/opensearch.job-scheduler.release-notes-1.2.0.0.md
@@ -1,0 +1,23 @@
+## Version 1.2.0.0 2021-11-05
+
+Compatible with OpenSearch 1.2.0
+
+### Features
+
+* Adds a delay parameter to the job scheduler ([#61](https://github.com/opensearch-project/job-scheduler/pull/61))
+
+### Bug Fixes
+
+* Fixes possible negative jitter ([#78](https://github.com/opensearch-project/job-scheduler/pull/78))
+
+### Infrastructure
+
+* Publish MD5 and SHA1 signatures ([#71](https://github.com/opensearch-project/job-scheduler/pull/71))
+
+### Documentation
+
+* Add release notes for 1.2.0.0 ([#84](https://github.com/opensearch-project/job-scheduler/pull/84))
+
+### Maintenance
+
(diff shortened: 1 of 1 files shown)


qreshi:
Looks like the link checker is failing on the install example in the [DEVELOPER_GUIDE](https://github.com/opensearch-project/job-scheduler/blob/main/DEVELOPER_GUIDE.md#install) because it's reading the file path as a link. Should we revert this to the previous example or is there a way to exempt it we prefer the current example?

release-notes/opensearch.job-scheduler.release-notes-1.2.0.0.md
```diff
@@ -0,0 +1,23 @@
+## Version 1.2.0.0 2021-11-05
+
+Compatible with OpenSearch 1.2.0
+
+### Features
+
+* Adds a delay parameter to the job scheduler ([#61](https://github.com/opensearch-project/job-scheduler/pull/61))
```
lezzago, line 7: This was part of 1.1 already

release-notes/opensearch.job-scheduler.release-notes-1.2.0.0.md
```diff
@@ -0,0 +1,23 @@
...
+
+### Bug Fixes
+
+* Fixes possible negative jitter ([#78](https://github.com/opensearch-project/job-scheduler/pull/78))
+
+### Infrastructure
+
+* Publish MD5 and SHA1 signatures ([#71](https://github.com/opensearch-project/job-scheduler/pull/71))
```
lezzago, line 15: This was part of 1.1 already

release-notes/opensearch.job-scheduler.release-notes-1.2.0.0.md
qreshi: It happened as patch after `1.1.0.0` so it's not going to be released until OS `1.1.1.0`. I've added it here since this will be the first release with the change. It should also be included in the `1.1.1.0` release notes since that's an addition to 1.1 that should be called out I think.

release-notes/opensearch.job-scheduler.release-notes-1.2.0.0.md
qreshi: I believe the same applies here as my comment for the delay parameter.

lezzago (approved):

PR #117, opened 2022-01-11
Update build.gradle to include job-scheduler jar when staging maven publications.

mch2:
Signed-off-by: Marc Handalian <handalm@amazon.com>

### Description
- updates build.gradle to ensure the base jar has the required pom fields for publishing to maven central.
- updates to include this jar when publishing to the local staging repository currently used by CI to pick up artifacts for promotion.
- Updates developer on spi jar to be opensearch.
- Updates build.sh script to be in sync with the script in opensearch-build.  The version of this script in this repo is a backup, a separate PR will be made to opensearch-build.
 
### Issues Resolved
closes #114 
 
### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #120, opened 2022-01-21
[1.2] updating build.sh script to be the same as main

### LINKED ISSUE #114: Publish job-scheduler JAR to maven, not only -spi
### Is your feature request related to a problem? Please describe

Currently the Anomaly Detection plugin needs the entire job-scheduler plugin and not just the job-scheduler-spi in order to run BWC testing during its build process. Currently in order to successfully pass our build workflow we have to build and place the JAR file for the job scheduler plugin version that is relevant for that build in our repo. For example when we version bumped to 1.2.2.0 as seen in this PR we had to remove our 1.2.1.0 job-scheduler JAR and replace it with 1.2.2.0 job-scheduler JAR. This process means we have to manually be building job-scheduler every-time we want to version bump. We have this open issue stating this problem: https://github.com/opensearch-project/anomaly-detection/issues/352

 

### Describe the solution you'd like

In order to avoid manually building the JAR each time we propose to also have the job-scheduler uploaded to maven each time there is a version bump so we can fetch the snapshot in order to complete our build process. Moving over to this process would mean however that we would be reliant on maven always having the most up to date job-scheduler snapshot as our build process would be relaying on that the same way it now relies on core and job-scheduler-spi for example.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

amitgalitz:
Signed-off-by: Amit Galitzky <amgalitz@amazon.com>

### Description
Updates build.sh script to be in sync with the script in the main branch. The build.sh was changed in this PR on main: #117. Only the build.sh change needed to be made in other branches. The version of this script in opensearch-build will be deleted once this change has been made to every supported branch. 
 
### Issues Resolved
#114 
 
### Check List

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -58,10 +58,10 @@ fi
 [[ "$SNAPSHOT" == "true" ]] && VERSION=$VERSION-SNAPSHOT
 [ -z "$OUTPUT" ] && OUTPUT=artifacts
 
-./gradlew publishShadowPublicationToMavenLocal -Dopensearch.version=$VERSION -Dbuild.snapshot=$SNAPSHOT
-./gradlew publishShadowPublicationToStagingRepository -Dopensearch.version=$VERSION -Dbuild.snapshot=$SNAPSHOT
-mkdir -p $OUTPUT/maven/org/opensearch/opensearch-job-scheduler-spi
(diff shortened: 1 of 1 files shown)


amitgalitz:
Seems like the test failure is a flaky test since it doesn't fail when I re-run with the same seed locally. Could someone from @opensearch-project/job-scheduler who has access to re-run the workflow try to re-run checks for this PR. Thank you!

dblock (approved):

bowenlan-amzn (approved):

PR #130, opened 2022-03-01
Run CI on JDK 8, 11, and 14.

dblock:
Signed-off-by: dblock <dblock@dblock.org>

### Description

Copy-paste from https://github.com/opensearch-project/common-utils/pull/121.

### Issues Resolved

Closes https://github.com/opensearch-project/job-scheduler/issues/103.
 
### Check List
  - [x] All tests pass


dblock:
# 17

Failing with 

```
A problem occurred evaluating root project 'opensearch-job-scheduler'.
> Failed to apply plugin class 'de.thetaphi.forbiddenapis.gradle.ForbiddenApisPlugin'.
   > Could not create plugin of type 'ForbiddenApisPlugin'.
      > Could not initialize class de.thetaphi.forbiddenapis.gradle.ForbiddenApisPlugin
```

# 8 

Failing with

```
A problem occurred evaluating root project 'opensearch-job-scheduler'.
> Failed to apply plugin 'opensearch.opensearchplugin'.
   > Could not create plugin of type 'PluginBuildPlugin'.
      > Could not generate a decorated class for type PluginBuildPlugin.
         > org/opensearch/gradle/plugin/PluginPropertiesExtension has been compiled by a more recent version of the Java Runtime (class file version 54.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

@reta Any ideas here? Seems like for 8 it means having to go and build OpenSearch gradle extensions with JDK 8, and for 17 I am not sure.

reta:
@dblock ahhh ... here we are,  Gradle 6.x does not support JDK-17 .... (in core we use JDK-11 to run Gradle but then JDK-17 is "injected" to run tests)

dblock:
> @dblock ahhh ... here we are, Gradle 6.x does not support JDK-17 .... (in core we use JDK-11 to run Gradle but then JDK-17 is "injected" to run tests)

Do you know how or have any ideas how to solve this for a plugin?

> and indeed, `build-tools` are requiring JDK-10 at least
> 
> ```
>   allprojects {
>     targetCompatibility = 10
>     sourceCompatibility = 10
>   }
> ```
> 
> so artifacts are not consumable by JDK-8

I could change that to 8, would that make sense/work?

reta:
> Do you know how or have any ideas how to solve this for a plugin?

We need Gradle 7+ but it won't work for 1.3 :( since the `build-tools` are not compatible, We could branch off `1.x` for the plugin and move `main` to 2.0, that would unblock JDK-17

> I could change that to 8, would that make sense/work?
We could try, yeah, I don't know the prehistory of JDK-10 compatibility sadly

dblock:
@reta would you give up on using JDK17 at runtime for 1.3.0? 

Note that we also found https://github.com/opensearch-project/security/issues/1653

reta:
@dblock  Gradle 7 is a major blocker here, going with 11 for 1.3 looks acceptable, moreover we know right now that some plugins are not compatible with JDK-17 (and not sure about JDK016 since the reflective access started to tighten there).

dblock:
> @dblock Gradle 7 is a major blocker here, going with 11 for 1.3 looks acceptable, moreover we know right now that some plugins are not compatible with JDK-17 (and not sure about JDK-16 since the reflective access started to tighten there).

Yep :( 

I've edited https://github.com/opensearch-project/opensearch-plugins/issues/64 to reflect what we think we want in 1.3.0.

reta:
@dblock found it [1], let me send the PR ...

[1] https://github.com/opensearch-project/OpenSearch/blob/main/buildSrc/src/main/java/org/opensearch/gradle/OpenSearchJavaPlugin.java#L176

dblock:
Updated this PR to build on 8, 11 and 14. Punting 17 for post 1.3.0 when we upgrade to Gradle 7 and drop JDK8.

reta:
@dblock JDK-8 is green, but for JDK-17 we may need a stretch with Gradle 7 ... I could submit the change but we need to branch off 2.0 or/and 1.x

dblock:
> @dblock JDK-8 is green, but for JDK-17 we may need a stretch with Gradle 7 ... I could submit the change but we need to branch off 2.0 or/and 1.x

We'll do this part of 2.0.

VachaShah (approved):

saratvemulapalli (approved):

reta (approved):

PR #151, opened 2022-03-19
Allow Job Scheduler Jobs to Lock Resources

downsrob:
### Current Locking system

The current locking system in the job-scheduler only allows for locking based on job_index_name and job_id. This extends the existing job scheduler LockService and LockModel to to be able to acquire locks on any arbitrary resource.

### Why extend the current locking service?

The motivation for the extension of the LockingService is for the [Index State Management Shrink Action.](https://github.com/opensearch-project/index-management/issues/40) The shrink action needs to place a lock on a specific node with enough space for the shrinking index so that no other Shrink job can simultaneously allocate shards to the same node. The previous job locking service is insufficient for this use case because it solely locks based on job_index_name and job_id, with the purpose of enabling users to keep separate scheduled runs of the same job from overlapping on a resource. This extension would allow for locking of specific resources across jobs per jobIndexName. 

For the Index Management use case, this locking system could be implemented in other ways, for example using the Cluster State. However, using the job scheduler lock index scales much better, and allows more services to use the index for more generic locks.

### Changes to LockModel

To allow for locking of arbitrary resources that extend across job_ids, this change adds an additional type of LockModel. This new lock type may be used through the resourceType and resource parameters added to the LockModel constructor. Using these parameters, the lock will be identified through a combination of the jobIndexName, the resourceType, and a hash of a map which should represent the resource.

### Issues Resolved
NA
 
### Check List
  - [X] All tests pass

### DIFF
--- a/spi/src/main/java/org/opensearch/jobscheduler/spi/utils/LockService.java
+++ b/spi/src/main/java/org/opensearch/jobscheduler/spi/utils/LockService.java
@@ -5,6 +5,7 @@
 
 package org.opensearch.jobscheduler.spi.utils;
 
+import org.opensearch.action.support.master.AcknowledgedResponse;
 import org.opensearch.jobscheduler.spi.JobExecutionContext;
 import org.opensearch.jobscheduler.spi.LockModel;
 import org.opensearch.jobscheduler.spi.ScheduledJobParameter;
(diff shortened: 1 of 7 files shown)


spi/src/main/java/org/opensearch/jobscheduler/spi/utils/LockService.java
```diff
@@ -68,6 +70,25 @@ private String lockMapping() {
...
+                            logger.error("Error updating lock index mapping.");
+                        }
+                        listener.onResponse(acknowledgedResponse.isAcknowledged());
+                    }
+
+                    @Override
+                    public void onFailure(Exception e) {
+                        listener.onFailure(e);
+                    }
+                }
```
thalurur, line 88: minor/nit: indentation seems off

PR #154, opened 2022-03-23
Increment to 1.4.0

ohltyler:
Signed-off-by: Tyler Ohlsen <ohltyler@amazon.com>

### Description
Increment to 1.4.0.
 
### Check List

### DIFF
--- a/.github/workflows/test-and-build-workflow.yml
+++ b/.github/workflows/test-and-build-workflow.yml
@@ -28,11 +28,11 @@ jobs:
       # job-scheduler
       - name: Build and Test
         run: |
-          ./gradlew build -Dopensearch.version=1.3.0-SNAPSHOT
+          ./gradlew build -Dopensearch.version=1.4.0-SNAPSHOT
 
       - name: Publish to Maven Local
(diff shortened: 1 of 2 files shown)


.github/workflows/test-and-build-workflow.yml
```diff
@@ -28,11 +28,11 @@ jobs:
       # job-scheduler
       - name: Build and Test
         run: |
-          ./gradlew build -Dopensearch.version=1.3.0-SNAPSHOT
```
dblock, line 31: How about we remove `-Dopensearch.version=` from here and let it pickup the default?

dblock (changes requested):

.github/workflows/test-and-build-workflow.yml
ohltyler: Good catch, I see that was already updated in `main`. Will update here.

amitgalitz (approved):

saratvemulapalli (approved):

dblock (approved):

PR #164, opened 2022-04-04
Enables locking on an arbitrary lockID

downsrob:
Signed-off-by: Clay Downs <downsrob@amazon.com>

### Description
This change allows for locking an arbitrary resource across jobs by repurposing the LockModel job_id to be any user provided lock ID String. The locks are still scoped to the specific job index provided.

Previously, locks could only be acquired by supplying the ScheduledJobParameter and JobExecutionContext. The job index name, job ID, and lock duration would then be extracted from those objects. This change adds a method to acquire a lock with any provided lock ID, and allows for explicitly providing the job index name and lock duration, so the ScheduledJobParameter and JobExecutionContext are no longer required.

### Issues Resolved
https://github.com/opensearch-project/job-scheduler/issues/166
 
### Check List
  - [X] All tests pass

### DIFF
--- a/spi/src/main/java/org/opensearch/jobscheduler/spi/LockModel.java
+++ b/spi/src/main/java/org/opensearch/jobscheduler/spi/LockModel.java
@@ -153,6 +153,23 @@ public final class LockModel implements ToXContentObject {
         );
     }
 
+    /*
+     * Parses the LockModel while also taking in the lockID, which will be the document ID of the lock being parsed.
+     */
+    public static LockModel parseWithID(final XContentParser parser, long seqNo, long primaryTerm, String lockId) throws IOException {
+        LockModel lockNoID = parse(parser, seqNo, primaryTerm);
+        return new LockModel(
+                lockNoID.jobIndexName,
+                lockNoID.jobId,
+                requireNonNull(lockId, "LockId cannot be when explicitly provided in the parser"),
+                lockNoID.lockTime,
+                lockNoID.lockDurationSeconds,
+                lockNoID.released,
+                seqNo,
+                primaryTerm
+        );
+    }
+
     @Override public XContentBuilder toXContent(final XContentBuilder builder, final Params params) throws IOException {
         builder.startObject()
             .field(JOB_INDEX_NAME, this.jobIndexName)
--- a/spi/src/main/java/org/opensearch/jobscheduler/spi/LockModel.java
+++ b/spi/src/main/java/org/opensearch/jobscheduler/spi/LockModel.java
@@ -43,7 +43,7 @@ public final class LockModel implements ToXContentObject {
      * @param primaryTerm primary term from OpenSearch document.
      */
     public LockModel(final LockModel copyLock, long seqNo, long primaryTerm) {
-        this(copyLock.jobIndexName, copyLock.jobId, copyLock.lockId, copyLock.lockTime, copyLock.lockDurationSeconds,
+        this(copyLock.jobIndexName, copyLock.jobId, copyLock.lockTime, copyLock.lockDurationSeconds,
             copyLock.released, seqNo, primaryTerm);
     }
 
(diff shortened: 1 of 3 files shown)


spi/src/main/java/org/opensearch/jobscheduler/spi/utils/LockService.java
```diff
@@ -106,6 +106,28 @@ public void acquireLock(final ScheduledJobParameter jobParameter,
...
+     *
+     * @param jobParameter a {@code ScheduledJobParameter} containing the lock duration.
+     * @param context a {@code JobExecutionContext} containing job index name and job id.
+     * @param lockId the unique Id for the lock. This should represent the resource that the lock is on, whether it be
+     *               a job, or some other arbitrary resource.
+     * @param listener an {@code ActionListener} that has onResponse and onFailure that is used to return the lock if it was acquired
+     *                 or else null. Passes {@code IllegalArgumentException} to onFailure if the {@code ScheduledJobParameter} does not
+     *                 have {@code LockDurationSeconds}.
+     */
+    public void acquireLockWithId(final ScheduledJobParameter jobParameter,
```
dbbaughe, line 125: If given a lockId by the caller, I think we should still prefix it with the job index name to preserve namespaces, otherwise if two different plugins using this end up using the same lockId at any point then they'll run into issues.

downsrob: I am happy to make this change if you want, but that behavior also enables locking resources across plugins. Any plugin that doesn't want this could just prefix their own id to namespace it.

dbbaughe: Is there a use case for that yet? If not then we can add in support for it when needed, but as of right now relying on the consumers to correctly namespace the lockId leaves room for the developers who are not aware of it to accidentally misuse and cause issues across plugins.

@thalurur Any thoughts on it?

thalurur: I don't see much benefit in enforcing a global lock since this is not a standard that all plugins need to follow and one plugin assuming that its locking globally using this mechanism might be really not enforcable and I don't think a plugin should assume that.

Keeping it at the a namespace always by job scheduler makes more sense since its contained to that use case namespace and if plugins want to further scope it down they can add more prefixes for the feature when creating lock Ids

spi/src/main/java/org/opensearch/jobscheduler/spi/LockModel.java
```diff
@@ -88,6 +94,18 @@ public LockModel(String jobIndexName, String jobId, Instant lockTime,
         this.primaryTerm = primaryTerm;
     }
 
+    public LockModel(String jobIndexName, String jobId, String lockId, Instant lockTime,
```
thalurur, line 97: why do we want both jobId and lockId?

downsrob: That's a good point. I think that the only reason to include the job ID would be if we wanted to delete the leftover locks after a job is deleted. If we instead moved to a periodic job deleting stale locks, then I think there is no point in having both. I'll make the change.

dbbaughe (approved):

spi/src/main/java/org/opensearch/jobscheduler/spi/LockModel.java
```diff
@@ -80,6 +80,8 @@ public LockModel(String jobIndexName, String jobId, Instant lockTime,
                      long lockDurationSeconds, boolean released, long seqNo, long primaryTerm) {
         this.lockId = jobIndexName + LOCK_ID_DELIMITR + jobId;
         this.jobIndexName = jobIndexName;
+        // The jobId parameter does not necessarily need to represent the id of a job scheduler job, as it is being used
```
thalurur, line 83: may be create an issue in repo to rename this completely to just lockId everywhere? so the paradigm is changed in job scheduler - cleaning up stale locks and any other refactor needed with this change

thalurur (approved):

spi/src/main/java/org/opensearch/jobscheduler/spi/LockModel.java
downsrob: Done!

PR #289, opened 2022-12-29
Communication mechanism for js

vibrantvarun:
### Description
Get Job Index and Get Job Type Api's to fetch job details for Job Scheduler
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-sdk-java/issues/274
 
### Check List
  - [x] All tests pass
- [x] New functionality has been documented.
  - [x] New functionality has javadoc added


src/main/java/org/opensearch/jobscheduler/JobSchedulerPlugin.java
```diff
@@ -5,6 +5,15 @@
 
 package org.opensearch.jobscheduler;
 
+import org.opensearch.action.ActionRequest;
+import org.opensearch.action.ActionResponse;
+import org.opensearch.client.node.NodeClient;
+import org.opensearch.cluster.node.DiscoveryNodes;
+import org.opensearch.common.settings.*;
```
dbwiddis, line 12: Don't use star imports.  (You probably want to disable them on your IDE)

src/main/java/org/opensearch/jobscheduler/JobSchedulerPlugin.java
```diff
@@ -20,36 +29,39 @@
...
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.io.IOException;
+import java.util.*;
```
dbwiddis, line 56: `rm * `
(Don't type this on the command line, though) ;)

src/test/java/org/opensearch/jobscheduler/transport/GetJobTypeTransportActionTests.java
```diff
@@ -0,0 +1,57 @@
...
+ 
+     }
+ 
+     @Test
+     public void testGetJobTypeTransportAction() {
+         action.doExecute(task, request, response);
+     }
+ }
+ 
+
```
dbwiddis, line 57: This file and several others in this PR don't end in a newline.  Spotless (if configured on this repo) should fix this (or we need to reconfigure this repo's spotless to match SDK for newline checking).  Spotless should also complain about the star imports :)

dbwiddis (changes requested):

src/main/java/org/opensearch/jobscheduler/JobSchedulerPlugin.java
vibrantvarun: done

src/main/java/org/opensearch/jobscheduler/JobSchedulerPlugin.java
vibrantvarun: done

src/test/java/org/opensearch/jobscheduler/transport/GetJobTypeTransportActionTests.java
vibrantvarun: done

src/main/java/org/opensearch/jobscheduler/utils/RestHandlerUtils.java
```diff
@@ -0,0 +1,83 @@
...
+
+public final class RestHandlerUtils {
+
+    private static final Logger logger = LogManager.getLogger(RestHandlerUtils.class);
+
+    private RestHandlerUtils() {}
+
+    /**
+     * Wrap action listener to avoid return verbose error message and wrong 500 error to user.
+     * Suggestion for exception handling in AD:
```
joshpalis, line 33: Why are comments about Anomaly Detection here?

src/test/java/org/opensearch/jobscheduler/transport/GetJobTypeTransportActionTests.java
```diff
@@ -0,0 +1,57 @@
...
+            public void onFailure(Exception e) {
+                // Assert.assertTrue(true);
+            }
+        };
+
+    }
+
+    @Test
+    public void testGetJobTypeTransportAction() {
+        action.doExecute(task, request, response);
```
dbwiddis, line 55: Should there be an assertion here?  Expected exception perhaps?

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobIndexActionTests.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Before
+    public void setUp() throws Exception {
+        super.setUp();
+        jobDetailsHashMap = new HashMap<>();
+        action = new RestGetJobIndexAction(jobDetailsHashMap);
+    }
+
+    @Test
+    public void getNameTests() {
```
joshpalis, line 46: lets follow unit test naming conventions, all test names should follow this format `test.......()`

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobTypeActionTests.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Test
+    public void getRoutes() {
+        List<RestHandler.Route> routes = action.routes();
+
+        Assert.assertEquals("/_plugins/_job_scheduler/_get/_job_type", routes.get(0).getPath());
+    }
+
+    @Test
+    public void prepareRequestTest() throws IOException {
```
joshpalis, line 59: change name to  `testPrepareRequest()`

src/test/java/org/opensearch/jobscheduler/transport/GetJobTypeTransportActionTests.java
```diff
@@ -0,0 +1,57 @@
...
+    public void setUp() throws Exception {
+        super.setUp();
+        action = new GetJobTypeTransportAction("", mock(TransportService.class), mock(ActionFilters.class));
+        request = new GetJobTypeRequest("demo_job_type", "extension_id");
+        task = mock(Task.class);
+        response = new ActionListener<GetJobDetailsResponse>() {
+            @Override
+            public void onResponse(GetJobDetailsResponse jobDetailsResponse) {
+                // onResponse will not be called as we do not have the AD index
+                Assert.assertEquals("success", jobDetailsResponse.getResponse());
```
dbwiddis, line 42: If this isn't supposed to be called why not do `fail("We do not have index so onResponse should not be called");`

joshpalis (commented):
Doing an initial pass

src/test/java/org/opensearch/jobscheduler/transport/GetJobIndexTransportActionTests.java
```diff
@@ -0,0 +1,57 @@
...
+            public void onFailure(Exception e) {
+                // Assert.assertTrue(true);
+            }
+        };
+
+    }
+
+    @Test
+    public void testGetJobIndexTransportAction() {
+        action.doExecute(task, request, response);
```
dbwiddis, line 55: Same questions here as `GetJobTypeTransportActionTests`

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobTypeActionTests.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Test
+    public void getNameTests() {
+        String name = action.getName();
+        Assert.assertEquals("get_job_type_action", name);
+    }
+
+    @Test
+    public void getRoutes() {
```
dbwiddis, line 52: should have "test" in the name somewhere.

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobTypeActionTests.java
```diff
@@ -0,0 +1,75 @@
...
+        List<RestHandler.Route> routes = action.routes();
+
+        Assert.assertEquals("/_plugins/_job_scheduler/_get/_job_type", routes.get(0).getPath());
```
dbwiddis, line 55: consider making this path a constant, it's used twice.

Also use the same constants as in the actual class: `String.format(Locale.ROOT, "%s/%s", JobSchedulerPlugin.JS_BASE_URI, "_get/_job_type"))`

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobIndexActionTests.java
```diff
@@ -0,0 +1,76 @@
...
+        String name = action.getName();
+        Assert.assertEquals("get_job_index_action", name);
+    }
+
+    @Test
+    public void getRoutes() {
+        List<RestHandler.Route> routes = action.routes();
+
+        Assert.assertEquals("/_plugins/_job_scheduler/_get/_job_index", routes.get(0).getPath());
```
dbwiddis, line 55: consider making this path a constant, it's used twice

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobIndexActionTests.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Test
+    public void getNameTests() {
+        String name = action.getName();
+        Assert.assertEquals("get_job_index_action", name);
+    }
+
+    @Test
+    public void getRoutes() {
```
dbwiddis, line 52: should have "test" in the name.

src/test/java/org/opensearch/jobscheduler/TestHelpers.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.opensearch.common.bytes.BytesReference;
+import org.opensearch.common.xcontent.XContentBuilder;
+import org.opensearch.common.xcontent.XContentType;
+
+import java.io.IOException;
+
+public class TestHelpers {
+
+    public static XContentBuilder builder() throws IOException {
```
dbwiddis, line 19: probably want "xContent" in the method name.

src/main/java/org/opensearch/jobscheduler/constant/CommonValue.java
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.jobscheduler.constant;
+
+public class CommonValue {
+
+    public static Integer NO_SCHEMA_VERSION = 0;
```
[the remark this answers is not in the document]
vibrantvarun: I took reference from AD. I can remove it

src/main/java/org/opensearch/jobscheduler/rest/RestGetJobTypeAction.java
```diff
@@ -0,0 +1,92 @@
...
+
+    private RestResponseListener<GetJobDetailsResponse> getJobTypeResponse(RestChannel channel, RestStatus status) {
+        return new RestResponseListener<>(channel) {
+            @Override
+            public RestResponse buildResponse(GetJobDetailsResponse response) throws Exception {
+                BytesRestResponse bytesRestResponse = new BytesRestResponse(
+                    status,
+                    response.toXContent(channel.newBuilder(), ToXContent.EMPTY_PARAMS)
+                );
+                return bytesRestResponse;
```
dbwiddis, line 88: Why create a temporary variable instead of just `return new BytesRestResponse( ... );`?

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobIndexActionTests.java
```diff
@@ -0,0 +1,76 @@
...
+    public void setUp() throws Exception {
+        super.setUp();
+        jobDetailsHashMap = new HashMap<>();
+        action = new RestGetJobIndexAction(jobDetailsHashMap);
+    }
+
+    @Test
+    public void getNameTests() {
+        String name = action.getName();
+        Assert.assertEquals("get_job_index_action", name);
```
dbwiddis, line 48: Consider using the constant in the RestGetJobIndexAction class.

src/main/java/org/opensearch/jobscheduler/rest/RestGetJobIndexAction.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.Locale;
+
+import static java.util.Arrays.asList;
+import static java.util.Collections.unmodifiableList;
+import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
+import static org.opensearch.rest.RestRequest.Method.PUT;
+
+public class RestGetJobIndexAction extends BaseRestHandler {
+
+    private static final String GET_JOB_INDEX_ACTION = "get_job_index_action";
```
dbwiddis, line 42: if you make this public you can use its value in tests

src/main/java/org/opensearch/jobscheduler/JobSchedulerPlugin.java
```diff
@@ -60,10 +79,12 @@ public class JobSchedulerPlugin extends Plugin implements ExtensiblePlugin {
     private LockService lockService;
     private Map<String, ScheduledJobProvider> indexToJobProviders;
     private Set<String> indicesToListen;
+    private HashMap<String, JobDetails> jobDetailsHashMap;
```
dbwiddis, line 82: Just declare this as a `Map` and remove `hash` from the name.

dbwiddis (changes requested):

src/test/java/org/opensearch/jobscheduler/transport/GetJobTypeTransportActionTests.java
vibrantvarun: No, it is a void method, the assertion is in listener.

src/test/java/org/opensearch/jobscheduler/transport/GetJobTypeTransportActionTests.java
vibrantvarun: This assertion is executed from the testPrepareRequest. The execution flow is when the call is made and then response is returned and handled by action listener

src/main/java/org/opensearch/jobscheduler/JobSchedulerPlugin.java
vibrantvarun: ACK

src/main/java/org/opensearch/jobscheduler/rest/RestGetJobIndexAction.java
vibrantvarun: ACk

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobIndexActionTests.java
vibrantvarun: ACK

src/main/java/org/opensearch/jobscheduler/rest/RestGetJobTypeAction.java
vibrantvarun: ACK

src/test/java/org/opensearch/jobscheduler/TestHelpers.java
vibrantvarun: ACK

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobIndexActionTests.java
vibrantvarun: ACK

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobTypeActionTests.java
vibrantvarun: ACK

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobTypeActionTests.java
vibrantvarun: ACK

src/test/java/org/opensearch/jobscheduler/transport/GetJobIndexTransportActionTests.java
vibrantvarun: ACK

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobTypeActionTests.java
vibrantvarun: ACK

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobIndexActionTests.java
vibrantvarun: ACK

src/main/java/org/opensearch/jobscheduler/utils/RestHandlerUtils.java
vibrantvarun: ACK

src/main/java/org/opensearch/jobscheduler/rest/RestGetJobIndexAction.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
+
+import static java.util.Arrays.asList;
+import static java.util.Collections.unmodifiableList;
+import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
+import static org.opensearch.rest.RestRequest.Method.PUT;
+
+public class RestGetJobIndexAction extends BaseRestHandler {
```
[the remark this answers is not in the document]
vibrantvarun: ACK

src/main/java/org/opensearch/jobscheduler/model/JobDetails.java
```diff
@@ -0,0 +1,161 @@
...
+import org.opensearch.common.xcontent.ToXContentObject;
+import org.opensearch.common.xcontent.XContentBuilder;
+import org.opensearch.common.xcontent.XContentParser;
+
+import java.io.IOException;
+import java.util.Objects;
+
+import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
+
+public class JobDetails implements ToXContentObject {
```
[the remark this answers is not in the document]
vibrantvarun: ACK

src/main/java/org/opensearch/jobscheduler/model/JobDetails.java
```diff
@@ -0,0 +1,161 @@
...
+
+    private String jobIndex;
+
+    private String jobType;
+
+    private String jobParamAction;
```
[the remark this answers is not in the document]
vibrantvarun: ACK

src/main/java/org/opensearch/jobscheduler/JobSchedulerPlugin.java
```diff
@@ -60,10 +79,12 @@ public class JobSchedulerPlugin extends Plugin implements ExtensiblePlugin {
     private LockService lockService;
     private Map<String, ScheduledJobProvider> indexToJobProviders;
     private Set<String> indicesToListen;
+    private Map<String, JobDetails> jobDetailsHashMap;
 
     public JobSchedulerPlugin() {
         this.indicesToListen = new HashSet<>();
         this.indexToJobProviders = new HashMap<>();
+        this.jobDetailsHashMap = new HashMap<>();
```
dbwiddis, line 87: (Naming) Would prefer a name that tells me what the key/value relationship is.  See the above map "index" to "JobProviders".   So I think `indexToJobDetails` is better here.

src/main/java/org/opensearch/jobscheduler/model/JobDetails.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        JobDetails that = (JobDetails) o;
+        return Objects.equals(jobIndex, that.jobIndex)
+            && Objects.equals(jobType, that.jobType)
+            && jobParserAction.equals(that.jobParserAction)
```
dbwiddis, line 138: Why are these last two a different format than the previous two (`Objects.equals()`)?   Also it looks like they can be `null` so this would throw an NPE, so you would want to use the `Objects` style.

src/main/java/org/opensearch/jobscheduler/transport/GetJobIndexRequest.java
```diff
@@ -0,0 +1,128 @@
...
+        return extensionId;
+    }
+
+    public void setExtensionId(String extensionId) {
+        this.extensionId = extensionId;
+    }
+
+    @Override
+    public ActionRequestValidationException validate() {
+        return null;
```
dbwiddis, line 96: This looks odd.  Is this an unimplemented TODO or is this the correct way to implement a no-op?

src/main/java/org/opensearch/jobscheduler/utils/RestHandlerUtils.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    public static boolean isProperExceptionToReturn(Throwable e) {
+        if (e == null) {
+            return false;
+        }
+        return e instanceof OpenSearchStatusException || e instanceof IndexNotFoundException || e instanceof InvalidIndexNameException;
+    }
+
+    private static String coalesceToEmpty(@Nullable String s) {
```
dbwiddis, line 76: This looks like pretty generic functionality. Is it available in another util class?  Seems it'd fit better in `org.opensearch.common.Strings`.  (Obviously can't do that in this PR but might be a nice feature request.)

src/test/java/org/opensearch/jobscheduler/TestHelpers.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.io.IOException;
+
+public class TestHelpers {
+
+    public static XContentBuilder xContentBuilder() throws IOException {
+        return XContentBuilder.builder(XContentType.JSON.xContent());
+    }
+
+    public static String xContentBuilderToString(XContentBuilder builder) {
```
dbwiddis, line 23: This one _is_ available in `Strings.toString(XContentBuilder builder)`. :-)

src/test/java/org/opensearch/jobscheduler/TestHelpers.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.opensearch.common.bytes.BytesReference;
+import org.opensearch.common.xcontent.XContentBuilder;
+import org.opensearch.common.xcontent.XContentType;
+
+import java.io.IOException;
+
+public class TestHelpers {
+
+    public static XContentBuilder xContentBuilder() throws IOException {
```
dbwiddis, line 19: This is available as `JsonXContent.contentBuilder()`.

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobIndexActionTests.java
```diff
@@ -0,0 +1,77 @@
...
+    @Before
+    public void setUp() throws Exception {
+        super.setUp();
+        jobDetailsHashMap = new HashMap<>();
+        action = new RestGetJobIndexAction(jobDetailsHashMap);
+    }
+
+    public void testGetNames() {
+        String name = action.getName();
+        Assert.assertEquals(action.GET_JOB_INDEX_ACTION, name);
```
dbwiddis, line 48: (Style) Almost everywhere else I see these assertions as static imports.

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobIndexActionTests.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    public void testGetRoutes() {
+        List<RestHandler.Route> routes = action.routes();
+
+        Assert.assertEquals(
+            String.format(Locale.ROOT, "%s/%s", JobSchedulerPlugin.JS_BASE_URI, "_get/_job_index"),
```
dbwiddis, line 55: Make this a constant or at least instance variable. You also use it below in withPath()`.

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobTypeActionTests.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public void testGetNames() {
+        String name = action.getName();
+        Assert.assertEquals(action.GET_JOB_TYPE_ACTION, name);
+    }
+
+    public void testGetRoutes() {
+        List<RestHandler.Route> routes = action.routes();
+
+        Assert.assertEquals(String.format(Locale.ROOT, "%s/%s", JobSchedulerPlugin.JS_BASE_URI, "_get/_job_type"), routes.get(0).getPath());
```
dbwiddis, line 54: Similar to previous comment: make the String an instance variable and use it in `.withPath()` below.

src/test/java/org/opensearch/jobscheduler/transport/GetJobTypeTransportActionTests.java
dbwiddis: I'm not sure I understand the comment on line 40, then.    Can we expect `onResponse()` to be called?  Can we expect `onFailure()` to not be called (in which case we should put a `fail()` call there.)

src/test/java/org/opensearch/jobscheduler/transport/GetJobTypeTransportActionTests.java
dbwiddis: You can still do an [`assertDoesNotThrow()`](https://junit.org/junit5/docs/5.8.2/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#assertDoesNotThrow(org.junit.jupiter.api.function.Executable)).

dbwiddis (changes requested):

src/main/java/org/opensearch/jobscheduler/transport/GetJobIndexRequest.java
vibrantvarun: It is a correct way to implement no-op. I have taken reference of this implementation from [GetAnomalyDetectorRequest](https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/transport/GetAnomalyDetectorRequest.java)

src/main/java/org/opensearch/jobscheduler/utils/RestHandlerUtils.java
vibrantvarun: ACK. Last month the same method was added in AD [RestHandlerUtils](https://github.com/opensearch-project/anomaly-detection/blob/feature/extensions/src/main/java/org/opensearch/ad/util/RestHandlerUtils.java)

src/test/java/org/opensearch/jobscheduler/TestHelpers.java
vibrantvarun: ACk

src/test/java/org/opensearch/jobscheduler/TestHelpers.java
vibrantvarun: ACK

src/main/java/org/opensearch/jobscheduler/utils/RestHandlerUtils.java
dbwiddis: Looks like the method _was_ in `Strings` until a month ago.  See https://github.com/opensearch-project/OpenSearch/issues/5664

OK for you to use the same `RestHandlerUtils` implementations as other plugins here for now.

src/test/java/org/opensearch/jobscheduler/transport/GetJobTypeTransportActionTests.java
vibrantvarun: ACk

src/test/java/org/opensearch/jobscheduler/transport/GetJobTypeTransportActionTests.java
vibrantvarun: Junit 5 is not supported in JS

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobTypeActionTests.java
vibrantvarun: ACK

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobIndexActionTests.java
vibrantvarun: ACK

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobIndexActionTests.java
vibrantvarun: ACK

src/main/java/org/opensearch/jobscheduler/model/JobDetails.java
vibrantvarun: ACK

src/main/java/org/opensearch/jobscheduler/JobSchedulerPlugin.java
vibrantvarun: ACK

joshpalis (dismissed):
LGTM

dbwiddis (dismissed):
Looks great!

joshpalis (approved):

joshpalis:
@saratvemulapalli We would like your review on this PR before we proceed with merging

src/main/java/org/opensearch/jobscheduler/constant/CommonValue.java
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.jobscheduler.constant;
+
+public class CommonValue {
+
+    public static String EXTERNAL_ACTION_PREFIX = "cluster:admin/opendistro/js/";
```
saratvemulapalli, line 13: Why `opendistro`?

vibrantvarun: I see, I am changing it to plugins. The reason to change as suggested by @saratvemulapalli is 

opendistro is part of ODFE: https://opendistro.github.io/for-elasticsearch/ which end of life.
AD security features were built in ODFE which is why it has legacy code
(https://amzn-aws.slack.com/archives/D03SBGET0BE/p1672437057031429)

joshpalis (dismissed):

src/main/java/org/opensearch/jobscheduler/transport/GetJobIndexAction.java
```diff
@@ -0,0 +1,25 @@
...
+
+import org.opensearch.action.ActionType;
+import org.opensearch.jobscheduler.constant.CommonValue;
+
+/**
+ * Get Job Index action
+ */
+public class GetJobIndexAction extends ActionType<GetJobDetailsResponse> {
+
+    public static final String NAME = CommonValue.EXTERNAL_ACTION_PREFIX + "get/jobIndex";
```
saratvemulapalli, line 19: nit: actions are preferred to be lower case.

src/main/java/org/opensearch/jobscheduler/transport/GetJobTypeAction.java
```diff
@@ -0,0 +1,25 @@
...
+
+import org.opensearch.action.ActionType;
+import org.opensearch.jobscheduler.constant.CommonValue;
+
+/**
+ * Get Job Type action
+ */
+public class GetJobTypeAction extends ActionType<GetJobDetailsResponse> {
+
+    public static final String NAME = CommonValue.EXTERNAL_ACTION_PREFIX + "get/jobType";
```
saratvemulapalli, line 19: same as the other action, lets use lower case, probably `job_type`

src/main/java/org/opensearch/jobscheduler/transport/GetJobTypeTransportAction.java
```diff
@@ -0,0 +1,42 @@
...
+public class GetJobTypeTransportAction extends HandledTransportAction<GetJobTypeRequest, GetJobDetailsResponse> {
+    private static final Logger LOG = LogManager.getLogger(GetJobTypeTransportAction.class);
+
+    @Inject
+    public GetJobTypeTransportAction(String actionName, TransportService transportService, ActionFilters actionFilters) {
+        super(GetJobTypeAction.NAME, transportService, actionFilters, GetJobTypeRequest::new);
+    }
+
+    @Override
+    protected void doExecute(Task task, GetJobTypeRequest request, ActionListener<GetJobDetailsResponse> actionListener) {
```
saratvemulapalli, line 35: What are we really processing in the transport layer? 
I see pretty much nothing we can't do at the Rest Layer.

dbwiddis (dismissed):

src/main/java/org/opensearch/jobscheduler/transport/GetJobTypeTransportAction.java
```diff
@@ -0,0 +1,42 @@
...
+        ActionListener<GetJobDetailsResponse> listener = wrapRestActionListener(
```
saratvemulapalli, line 36: Trying to understand, `wrapRestActionListener` wrapper should have been handling all of that in the Rest Layer. 
Why do we need this in transport?

saratvemulapalli (commented):
I see a lot of code we've added is going to transport but we aren't really adding value sending it over. 

I understand we've discussed this for security. But we are not processing any information and using security user context.

src/main/java/org/opensearch/jobscheduler/model/JobDetails.java
```diff
@@ -0,0 +1,176 @@
...
+
+    /**
+     * jobParser action to trigger the response back to the extension.
+     */
+    private String jobParserAction;
+
+    /**
+     * jobRunner action to trigger the response back to the extension.
+     */
+    private String jobRunnerAction;
```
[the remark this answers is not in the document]
vibrantvarun: There will be 2 callback methods. One to retrieve ScheduleJobParser and other would be ScheduleJobRunner

src/main/java/org/opensearch/jobscheduler/model/JobDetails.java
```diff
@@ -0,0 +1,176 @@
...
+        this.jobIndex = jobIndex;
+        this.jobType = jobType;
+        this.jobParserAction = jobParserAction;
+        this.jobRunnerAction = jobRunnerAction;
+    }
+
+    @Override
+    public XContentBuilder toXContent(XContentBuilder builder, ToXContent.Params params) throws IOException {
+        XContentBuilder xContentBuilder = builder.startObject();
+        if (jobIndex != null) {
```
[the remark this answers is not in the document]
vibrantvarun: yes we do have to check if they are null or not as while testing if we provide no value then it should handle that test case gracefully.

src/main/java/org/opensearch/jobscheduler/model/JobDetails.java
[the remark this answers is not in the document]
saratvemulapalli: Hmm why though, could you help me understand the flow? May be im missing the pieces to pull the puzzle together?

src/main/java/org/opensearch/jobscheduler/model/JobDetails.java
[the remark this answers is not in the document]
vibrantvarun: When the api call will be made then the job index will never be null.

src/main/java/org/opensearch/jobscheduler/transport/GetJobIndexAction.java
vibrantvarun: can you please elaborate it with an example?

vibrantvarun: ACK

src/main/java/org/opensearch/jobscheduler/transport/GetJobTypeAction.java
vibrantvarun: ACK

src/main/java/org/opensearch/jobscheduler/transport/GetJobTypeTransportAction.java
vibrantvarun: To return the response as 
{
"response":"success"
}

we need to implement the above logic in transport as layer. As the prepare request method return type is RestChannelConsumer, therefore we need to implement the logic for the same in transport layer. The response cannot be returned from Rest layer only

src/main/java/org/opensearch/jobscheduler/transport/GetJobTypeTransportAction.java
vibrantvarun: Without this I cannot return a response from Restchannel consumer.
The success response is sent in listener.OnResponse Method. The definition of onResponse is defined on rest level.

src/main/java/org/opensearch/jobscheduler/model/JobDetails.java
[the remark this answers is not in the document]
vibrantvarun: In JobSweeper. java sweep method

` ScheduledJobProvider provider = this.indexToProviders.get(shardId.getIndexName());
                    XContentParser parser = XContentHelper.createParser(
                        this.xContentRegistry,
                        LoggingDeprecationHandler.INSTANCE,
                        jobSource,
                        XContentType.JSON
                    );
                    ScheduledJobParameter jobParameter = provider.getJobParser().parse(parser, docId, jobDocVersion);
                    if (jobParameter == null) {
                        // allow parser to return null, which means this is not a scheduled job document.
                        return null;
                    }
                    ScheduledJobRunner jobRunner = this.indexToProviders.get(shardId.getIndexName()).getJobRunner();
                    if (jobParameter.isEnabled()) {
                        this.scheduler.schedule(shardId.getIndexName(), docId, jobParameter, jobRunner, jobDocVersion, jitterLimit);
                    }
                    return jobDocVersion;`

src/main/java/org/opensearch/jobscheduler/transport/GetJobTypeTransportAction.java
saratvemulapalli: I understand we have to respond with a success message, but what are we processing the transport layer from the request params? Pretty much nothing. 

We could just send the response from Rest `prepareRequest` as:

```
return new RestChannelConsumer() {
            
            final String mode = request.param("mode","strict");

            @Override
            public void accept(RestChannel channel) throws Exception {
                RestStatus restStatus = RestStatus.OK;
                BytesRestResponse response = null;
                try {
                    String status = "Sucess";
                    String message = "OK";
                    builder.startObject();

    
                    builder.field("message", message);
                    builder.field("mode", mode);
                    builder.field("status", status);
                    builder.endObject();
                    response = new BytesRestResponse(restStatus, builder);
                } finally {
                    builder.close();
                }
                channel.sendResponse(response);
            }
        };
```

src/main/java/org/opensearch/jobscheduler/model/JobDetails.java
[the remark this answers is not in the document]
joshpalis: @saratvemulapalli We'll need to transfer the bytesReference (jobDocVersion) to AD to parse the `ScheduledJobParameter` (AnomalyDetectorJob) before we are able to schedule the job and invoke the Job Runner.

saratvemulapalli (dismissed):
I see couple of problems:

- Going down to transport layer is really not needed, as we aren't processing anything but just to ack a response. 
- The information of `indexToJobDetails` map is going to become inconsistent when multiple clusters are in play. Which means, JS on node1 will respond with different information than JS on node2 and notification to AD extension is not called by each JS node. To solve this problem, plugins use metadata indices to put information and read from for consistency. 
Lets put this information into `.opendistro-job-scheduler` which is the existing metadata index and use this index to mutate/respond to the get calls. 

Also take a look if existing JS metadata index is a systemindex, if it is we need an admin client to put into the index. We can explore couple of options:
- Use transport layer and use super admin permissions to mutate the index.
- Or create another index just for JS extensions support which is not a system index.

src/main/java/org/opensearch/jobscheduler/model/JobDetails.java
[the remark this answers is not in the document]
vibrantvarun: ACK

src/main/java/org/opensearch/jobscheduler/transport/GetJobTypeTransportAction.java
vibrantvarun: Implemented the above approach in the latest PR

src/test/java/org/opensearch/jobscheduler/utils/JobDetailsServiceIT.java
```diff
@@ -0,0 +1,142 @@
...
+                    latch.countDown();
+                }, exception -> { fail(exception.getMessage()); }));
+
+                jobDetailsService.deleteJobDetailsForExtension("sample-extension", ActionListener.wrap(response -> {
+                    assertTrue(response);
+                    latch.countDown();
+                }, exception -> { fail(exception.getMessage()); }));
+            }, exception -> { fail(exception.getMessage()); })
+        );
+        latch.await(5L, TimeUnit.SECONDS);
```
joshpalis, line 59: lets use the timevalue constant here as well

src/test/java/org/opensearch/jobscheduler/utils/JobDetailsServiceIT.java
```diff
@@ -0,0 +1,142 @@
...
+                        deleted -> { assertTrue("Failed to delete job details for extension.", deleted); },
+                        exception -> fail(exception.getMessage())
+                    )
+                );
+            } else {
+                fail("Failed to job details for extension");
+            }
+
+        }, exception -> fail(exception.getMessage())));
+        latch.await(5L, TimeUnit.SECONDS);
```
joshpalis, line 139: and here

build.gradle
```diff
@@ -42,7 +42,7 @@ ext {
     noticeFile = rootProject.file('NOTICE')
 }
 
-licenseHeaders.enabled = true
```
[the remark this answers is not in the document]
vibrantvarun: I have made this condition false because it was throwing and error for no license in a json file. opensearch_plugins_job_details.json

src/main/java/org/opensearch/jobscheduler/model/JobDetails.java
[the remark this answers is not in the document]
saratvemulapalli: With extensions, the parser is not really passed to JS to run, probably lets name it get job param as an API as we discussed offline.

src/test/java/org/opensearch/jobscheduler/utils/JobDetailsServiceIT.java
vibrantvarun: ACk

src/test/java/org/opensearch/jobscheduler/utils/JobDetailsServiceIT.java
vibrantvarun: ACk

src/main/java/org/opensearch/jobscheduler/transport/GetJobIndexRequest.java
```diff
@@ -0,0 +1,130 @@
...
+    private static String jobParameterAction;
+
+    private static String jobRunnerAction;
+
+    private static String extensionId;
+
+    public static final String JOB_INDEX = "job_index";
+
+    public static final String EXTENSION_ID = "extension_id";
+    private static final String JOB_PARAMETER_ACTION = "job_parser_action";
```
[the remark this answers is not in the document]
vibrantvarun: ACk

build.gradle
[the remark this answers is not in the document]
vibrantvarun: Condition reverted now

joshpalis (commented):
Doing an initial pass

src/test/java/org/opensearch/jobscheduler/ODFERestTestCase.java
```diff
@@ -0,0 +1,217 @@
...
+        return isHttps() ? "https" : "http";
+    }
+
+    @Override
+    protected Settings restAdminSettings() {
+        return Settings.builder()
+            // disable the warning exception for admin client since it's only used for cleanup.
+            .put("strictDeprecationMode", false)
+            .put("http.port", 9200)
+            // @anomaly-detection.create-detector Commented this code until we have support of Common Utils for extensibility
```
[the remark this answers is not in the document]
vibrantvarun: done

src/test/java/org/opensearch/jobscheduler/ODFERestTestCase.java
```diff
@@ -0,0 +1,217 @@
...
+        Response response = adminClient().performRequest(request);
+        return RestStatus.OK.getStatus() == response.getStatusLine().getStatusCode();
+    }
+
+    @Override
+    protected RestClient buildClient(Settings settings, HttpHost[] hosts) throws IOException {
+        boolean strictDeprecationMode = settings.getAsBoolean("strictDeprecationMode", true);
+        RestClientBuilder builder = RestClient.builder(hosts);
+        if (isHttps()) {
+            // @anomaly-detection.create-detector Commented this code until we have support of Common Utils for extensibility
```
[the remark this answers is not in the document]
vibrantvarun: done

src/test/java/org/opensearch/jobscheduler/multinode/GetJobDetailsMultiNodeRestIT.java
```diff
@@ -0,0 +1,65 @@
...
+                null
+            );
+
+            String jobIndex = validateResponseAndGetJobIndex(entityAsMap(response1));
+            Assert.assertEquals(expectedJobIndex, jobIndex);
+        }
+    }
+
+    private String validateResponseAndGetJobIndex(Map<String, Object> responseMap) {
+        Assert.assertEquals("success", responseMap.get("response"));
```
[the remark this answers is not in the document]
vibrantvarun: done

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobIndexActionTests.java
```diff
@@ -0,0 +1,86 @@
...
+                null,
+                "sample-job-parameter-action",
+                "sample-runner-name",
+                "sample-extension",
+                JobDetailsService.JobDetailsRequestType.JOB_INDEX,
+                ActionListener.wrap(response -> {}, exception -> {})
+            );
+        action.prepareRequest(request, Mockito.mock(NodeClient.class));
+
+        assertThat(channel.responses().get(), equalTo(0));
```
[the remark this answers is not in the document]
vibrantvarun: done

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobIndexActionTests.java
```diff
@@ -0,0 +1,86 @@
...
+        assertThat(channel.errors().get(), equalTo(0));
```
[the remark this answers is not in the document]
vibrantvarun: done

src/test/java/org/opensearch/jobscheduler/rest/RestGetJobTypeActionTests.java
```diff
@@ -0,0 +1,85 @@
...
+                "sample-job-type",
+                null,
+                null,
+                "sample-extension",
+                JobDetailsService.JobDetailsRequestType.JOB_INDEX,
+                ActionListener.wrap(response -> {}, exception -> {})
+            );
+        action.prepareRequest(request, Mockito.mock(NodeClient.class));
+
+        assertThat(channel.responses().get(), CoreMatchers.equalTo(0));
```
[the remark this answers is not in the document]
vibrantvarun: done

src/main/java/org/opensearch/jobscheduler/utils/JobDetailsService.java
```diff
@@ -0,0 +1,306 @@
...
+        final String extensionId,
+        final JobDetailsRequestType requestType,
+        ActionListener<JobDetails> listener
+    ) {
+        boolean isJobIndexRequest;
+        if (requestType.JOB_INDEX == requestType) {
+            isJobIndexRequest = true;
+            if (jobIndexName == null) {
+                listener.onFailure(new IllegalArgumentException("Job Index Name should not be null"));
+            } else if (jobParameterActionName == null) {
```
dbwiddis, line 111: Since none of these can be null, seems we should include all of them in one exception.  So a single if check whether any are null, and an IAE listing which ones are null.

src/main/java/org/opensearch/jobscheduler/utils/JobDetailsService.java
```diff
@@ -0,0 +1,306 @@
...
+                listener.onFailure(new IllegalArgumentException("Job Parameter Action Name should not be null"));
+            } else if (jobRunnerActionName == null) {
+                listener.onFailure(new IllegalArgumentException("Job Runner Action Name should not be null"));
+            }
+        } else {
+            isJobIndexRequest = false;
```
dbwiddis, line 117: Interesting that we handle this separately.  Seems that this should never really be null and therefore trapped/caught earlier, while the other three might be.   Any chance we could use `Optional<String>` instead of null?    Then instead of null checks (see above) we could use `anyOf()` or `allOf()` or similar checks.

src/main/java/org/opensearch/jobscheduler/utils/JobDetailsService.java
```diff
@@ -0,0 +1,306 @@
...
+            if (jobTypeName == null) {
+                listener.onFailure(new IllegalArgumentException("Job Type Name should not be null"));
+            }
+        }
+        if (extensionId == null) {
+            listener.onFailure(new IllegalArgumentException("Extension Id should not be null"));
```
dbwiddis, line 123: "Should" or "must" have formalized meaning in standards.  Must is absolutely a requirement (error if wrong) while should is a recommendation.  These error messages _should_ conform.

src/test/java/org/opensearch/jobscheduler/ODFERestTestCase.java
```diff
@@ -0,0 +1,217 @@
...
+    protected String getProtocol() {
+        return isHttps() ? "https" : "http";
+    }
+
+    @Override
+    protected Settings restAdminSettings() {
+        return Settings.builder()
+            // disable the warning exception for admin client since it's only used for cleanup.
+            .put("strictDeprecationMode", false)
+            .put("http.port", 9200)
```
dbwiddis: Can we make the 9200 a constant at the top of the class?

src/test/java/org/opensearch/jobscheduler/multinode/GetJobDetailsMultiNodeRestIT.java
```diff
@@ -0,0 +1,65 @@
...
+     */
+    public void testGetJobDetailsRestAPI() throws Exception {
+
+        Response response = TestHelpers.makeRequest(
+            client(),
+            "PUT",
+            TestHelpers.GET_JOB_INDEX_BASE_DETECTORS_URI,
+            ImmutableMap.of(),
+            TestHelpers.toHttpEntity(
+                "{\"job_index\":\"demo_job_index\",\"job_parameter_action\":\"demo_parameter\",\"job_runner_action\":\"demo_runner\",\"extension_id\":\"sample_extension\"}"
```
dbwiddis, line 36: I'm curious in a multi-node IT whether we may get name collisions with constant named indices.  Have we considered using random strings here?

src/main/java/org/opensearch/jobscheduler/utils/JobDetailsService.java
```diff
@@ -0,0 +1,306 @@
...
+            logger.error("IOException occurred creating job details for extension id " + extensionId, e);
+            listener.onResponse(null);
+        }
+    }
+
+    /**
+     *
+     * @param extensionId unique id to find the job details document in the index
+     * @param listener an {@code ActionListener} that has onResponse and onFailure that is used to return the job details if it was found
+     *      *                 or else null.
```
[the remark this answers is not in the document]
vibrantvarun: Ack

src/main/java/org/opensearch/jobscheduler/utils/JobDetailsService.java
```diff
@@ -0,0 +1,306 @@
...
+            logger.error("Exception occurred finding job details for extension id " + extensionId, exception);
+            listener.onFailure(exception);
+        }));
+    }
+
+    /**
+     *
+     * @param extensionId unique id to find and delete the job details document in the index
+     * @param listener an {@code ActionListener} that has onResponse and onFailure that is used to return the job details if it was deleted
+     *      *                 or else null.
```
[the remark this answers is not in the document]
vibrantvarun: Ack

src/main/java/org/opensearch/jobscheduler/utils/JobDetailsService.java
```diff
@@ -0,0 +1,306 @@
...
+            }
+        }));
+    }
+
+    /**
+     *
+     * @param updateJobDetails update job details object entry
+     * @param extensionId unique id to find and update the corresponding document mapped to it
+     * @param listener an {@code ActionListener} that has onResponse and onFailure that is used to return the job details if it was updated
+     *      *                 or else null.
```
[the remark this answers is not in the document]
vibrantvarun: Ack

src/main/java/org/opensearch/jobscheduler/utils/JobDetailsService.java
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    public boolean jobDetailsIndexExist() {
+        return clusterService.state().routingTable().hasIndex(JOB_DETAILS_INDEX_NAME);
+    }
+
+    /**
+     *
+     * @param listener an {@code ActionListener} that has onResponse and onFailure that is used to return the job details index if it was created
+     *      *                 or else null.
```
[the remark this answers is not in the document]
vibrantvarun: Ack

src/test/java/org/opensearch/jobscheduler/ODFERestTestCase.java
vibrantvarun: done

src/test/java/org/opensearch/jobscheduler/multinode/GetJobDetailsMultiNodeRestIT.java
vibrantvarun: ACK

src/main/java/org/opensearch/jobscheduler/utils/JobDetailsService.java
vibrantvarun: done

src/main/java/org/opensearch/jobscheduler/utils/JobDetailsService.java
vibrantvarun: done

src/main/java/org/opensearch/jobscheduler/rest/RestGetJobIndexAction.java
```diff
@@ -0,0 +1,140 @@
...
+                    restResponse[0] = "failed";
+                    latch.countDown();
+                    logger.info("could not process job index", e);
+                }
+            }
+        );
+
+        try {
+            latch.await(JobDetailsService.TIME_OUT_FOR_LATCH, TimeUnit.SECONDS);
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
joshpalis: I agree, @vibrantvarun perhaps we could use multiple catches to catch specific exceptions, with specific error messages

joshpalis (approved):
LGTM other than the exception handling for the timeouts.

src/main/java/org/opensearch/jobscheduler/rest/RestGetJobIndexAction.java
```diff
@@ -0,0 +1,135 @@
...
+                @Override
+                public void onResponse(JobDetails jobDetails) {
+                    jobDetailsResponseHolder[0] = jobDetails;
+                    inProgressFuture.complete(jobDetailsResponseHolder);
+                }
+
+                @Override
+                public void onFailure(Exception e) {
+                    logger.info("could not process job index", e);
+                    jobDetailsResponseHolder[0] = null;
```
dbwiddis, line 99: This isn't necessary as the value is already null from the array instantiation.  It's reasonable from a readability standpoint to add a comment here indicating this remains null.

src/main/java/org/opensearch/jobscheduler/rest/RestGetJobIndexAction.java
```diff
@@ -0,0 +1,135 @@
...
+                    inProgressFuture.completeExceptionally(e);
+                }
+            }
+        );
+
+        try {
+            inProgressFuture.orTimeout(JobDetailsService.TIME_OUT_FOR_REQUEST, TimeUnit.SECONDS);
+            inProgressFuture.get(JobDetailsService.TIME_OUT_FOR_REQUEST, TimeUnit.SECONDS);
```
dbwiddis, line 107: You don't need the timeout on both, they'll run sequentially.  Suggest you do
```java
inProgressFuture.orTimeout(JobDetailsService.TIME_OUT_FOR_REQUEST, TimeUnit.SECONDS).get();
```
... or `.join()` which throws different types of exceptions, doesn't throw checked exceptions, and may be preferred depending on how you choose to handle the exceptions.  Specifically, `.get()` requires you to catch `InterruptedException` and `ExecutionException`) while `.join()` does not.

src/main/java/org/opensearch/jobscheduler/rest/RestGetJobIndexAction.java
```diff
@@ -0,0 +1,135 @@
...
+        } catch (Exception e) {
+            logger.info("Time Limit Exceeded due to exception ", e);
+        }
+
+        return channel -> {
+            XContentBuilder builder = channel.newBuilder();
+            RestStatus restStatus = RestStatus.OK;
+            String restResponse = jobDetailsResponseHolder[0] != null ? "success" : "failed";
```
dbwiddis, line 115: (Style/Naming) So this isn't confused with the `bytesRestResponse`, suggest either `restResponseValue` or `restResponseString` or similar here.

src/main/java/org/opensearch/jobscheduler/rest/RestGetJobTypeAction.java
```diff
@@ -0,0 +1,131 @@
...
+                    logger.info("could not process job type", e);
+                    jobDetailsResponseHolder[0] = null;
+                    inProgressFuture.complete(jobDetailsResponseHolder);
+                }
+            }
+        );
+
+        try {
+            inProgressFuture.orTimeout(JobDetailsService.TIME_OUT_FOR_REQUEST, TimeUnit.SECONDS);
+            inProgressFuture.get();
```
dbwiddis, line 103: This is the correct sequential way to handle (or you could chain `.get()` on the previous line, or `.join()`.  Same comments regarding handling timeout exception differently than an exception during execution.

src/main/java/org/opensearch/jobscheduler/rest/RestGetJobTypeAction.java
```diff
@@ -0,0 +1,131 @@
...
+        } catch (Exception e) {
+            logger.info("Time Limit Exceeded due to exception", e);
+        }
+
+        return channel -> {
+            XContentBuilder builder = channel.newBuilder();
+            RestStatus restStatus = RestStatus.OK;
+            String restResponse = jobDetailsResponseHolder[0] != null ? "success" : "failed";
```
dbwiddis, line 111: Same comments regarding naming.

src/main/java/org/opensearch/jobscheduler/utils/JobDetailsService.java
```diff
@@ -0,0 +1,309 @@
...
+                }
+            }
+        }, exception -> {
+            logger.error("Exception occurred finding job details for extension id " + extensionId, exception);
+            listener.onFailure(exception);
+        }));
+    }
+
+    /**
+     *
```
dbwiddis, line 231: Need a description of what the method actually does.

src/main/java/org/opensearch/jobscheduler/model/JobDetails.java
```diff
@@ -0,0 +1,180 @@
...
+        if (jobType != null) {
+            xContentBuilder.field(JOB_TYPE, jobType);
+        }
+        if (jobParameterAction != null) {
+            xContentBuilder.field(JOB_PARAMETER_ACTION, jobParameterAction);
+        }
+        if (jobRunnerAction != null) {
+            xContentBuilder.field(JOB_RUNNER_ACTION, jobRunnerAction);
+        }
+        return xContentBuilder.endObject();
```
dbwiddis, line 75: Nulls are dangerous and usually best avoided.  Personal preference is to try to use `Optional` instead.  I can be convinced to stick with null but will ask for it to be amply documented in that case.

Looking at all the classes in this PR, it seems that we always require `jobType` to be non-null, but the other values are allowed to be null.  That isn't obvious from looking at this class in isolation.  I suggest:
1. Test the value of `jobType` before returning here, and throw an appropriate parsing exception here if it is null.
2. Add the `@Nullable` annotation to the getters for the other three fields.
3. Add the `@Nullable` annotation to each of the (allowed to be null) parameters on the constructor.
4. Add a javadoc somewhere (perhaps the class itself, or this parsing method) indicating the allowed/disallowed nulls.

dbwiddis (changes requested):

src/main/java/org/opensearch/jobscheduler/rest/RestGetJobIndexAction.java
vibrantvarun: done

src/main/java/org/opensearch/jobscheduler/rest/RestGetJobIndexAction.java
vibrantvarun: done

src/main/java/org/opensearch/jobscheduler/rest/RestGetJobIndexAction.java
vibrantvarun: done

src/main/java/org/opensearch/jobscheduler/utils/JobDetailsService.java
vibrantvarun: done

src/main/java/org/opensearch/jobscheduler/model/JobDetails.java
vibrantvarun: done

src/main/java/org/opensearch/jobscheduler/rest/RestGetJobTypeAction.java
vibrantvarun: done

src/main/java/org/opensearch/jobscheduler/rest/RestGetJobTypeAction.java
vibrantvarun: done

dbwiddis (approved):

joshpalis (approved):

PR #294, opened 2023-01-04
Increasing thread sleep to 100,000

joshpalis:
Signed-off-by: Joshua Palis <jpalis@amazon.com>

### Description
There has been an increase of BWC test failures now that the windows CI workflow has been added. The source of the issue is the `createBasicWatcherJob()` method [here](https://github.com/opensearch-project/job-scheduler/blob/140ffd3d48f928db78fa9fb23242e8b50282f7aa/sample-extension-plugin/src/test/java/org/opensearch/jobscheduler/sampleextension/bwc/JobSchedulerBackwardsCompatibilityIT.java#L143) that counts the number of records within a job index after the job has been executed. The job is configured to execute every minute, and after waiting for 80,000 ms, the test attempts to confirm that there is only 1 record within the index. This error is not reproducible locally, and I suspect that this may be a transient fault. In order to mitigate this issue, increasing the thread sleep to 100,000

Example build failure : https://github.com/opensearch-project/job-scheduler/actions/runs/3803814253/jobs/6470491899#step:4:1189

 
### Check List
  - [ ] All tests pass


saratvemulapalli (approved):

dbwiddis (approved):

qreshi:
This seems a little strange. The helper method that indexes the document seems to be calling `refresh` already. It shouldn't even be taking `80,000ms` for the document to be searchable. Are we sure there isn't something else impeding on the assertion (like the tests leaking and causing the index to be cleaned up)?

PR #299, opened 2023-01-10
[Extensions] Synchronize opensearch-plugin-job-details with JobDetailsService map

joshpalis:
### Description
Adds `JobDetailsService` as an indexOperationListener for index `.opensearch-plugin-job-details`. This allows us to update the internal JobDetails map with the entries indexed to the job details meta data index. Additionally, the `indicesToListen` set is updated with the job index names post-index, configuring `onIndexModule` to add the `JobSweeper` as an indexOperationListener once an extension creates their job index
 
### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-sdk-java/issues/309
 
### Check List
  - [x] All tests pass


vibrantvarun:
Once the sanity test case for the workflow is added in the PR. LGTM!

src/main/java/org/opensearch/jobscheduler/utils/JobDetailsService.java
```diff
@@ -33,31 +33,89 @@
...
+        this.indicesToListen = indicesToListen;
+        this.indexToJobDetails = new ConcurrentHashMap<>();
     }
 
     public boolean jobDetailsIndexExist() {
         return clusterService.state().routingTable().hasIndex(JOB_DETAILS_INDEX_NAME);
     }
 
+    public ConcurrentMap<String, JobDetails> getIndexToJobDetails() {
+        return this.indexToJobDetails;
```
vibrantvarun, line 72: Implement the return of object of indexToJobdetails in below manner.

private static class InstanceHolder {
    private static final YourObject instance = new YourObject();
}

public static YourObject getInstance() {
    return InstanceHolder.instance;
}
This solution takes advantage of the Java memory model's guarantees about class initialization to ensure thread safety. Each class can only be loaded once, and it will only be loaded when it is needed. That means that the first time getInstance is called, InstanceHolder will be loaded and instance will be created, and since this is controlled by ClassLoaders, no additional synchronization is necessary.

In the case above it would look like 

JobDetailsService.indexToJobDetails

joshpalis: Sure I'll go ahead and add this, I'll keep the method name `getIndexToJobDetails`, appending Map isn't necessary

joshpalis:
BWC tests are failing since JS main BWC tests use JS 2.5.0, which is no longer wire-compatible with JS 3.x. Currently waiting for OpenSearch `2.6.0-SNAPSHOT` to be available in sonatype before merging https://github.com/opensearch-project/job-scheduler/pull/297. Upon merging, I'll update the JS main BWC tests to use 2.6.0. This should resolve the failing checks

src/test/java/org/opensearch/jobscheduler/utils/JobDetailsServiceIT.java
```diff
@@ -137,4 +142,57 @@ public void testDeleteNonExistingJobDetails() throws ExecutionException, Interru
...
     }
 
+    public void testUpdateIndexToJobDetails() throws ExecutionException, InterruptedException, TimeoutException {
+        String extensionId = "sample-extension";
+        CompletableFuture<Boolean> inProgressFuture = new CompletableFuture<>();
+        JobDetailsService jobDetailsService = new JobDetailsService(client(), this.clusterService, this.indicesToListen);
+
+        // Index Job Index name, actions, for extensionID
+        jobDetailsService.processJobDetailsForExtensionId(
+            "sample-job-index",
```
dbwiddis, line 152: Suggest making these strings local variables (e.g., `expectedJobIndex`) as they are reused:
 - lines 152-155 and 162
 - lines 169-172
 - should use them at lines 183-186

src/test/java/org/opensearch/jobscheduler/utils/JobDetailsServiceIT.java
```diff
@@ -137,4 +142,57 @@ public void testDeleteNonExistingJobDetails() throws ExecutionException, Interru
...
+                        // We'll have to invoke updateIndexToJobDetails as jobDetailsService is added as an indexOperationListener
+                        // onIndexModule
+                        jobDetailsService.updateIndexToJobDetails(extensionId, jobDetails);
+
+                        // Ensure indicesToListen is updated
+                        assertTrue(this.indicesToListen.contains(jobDetails.getJobIndex()));
+
+                        // Ensure indexToJobDetails is updated
+                        JobDetails entry = jobDetailsService.getIndexToJobDetails().get(extensionId);
+                        assertEquals(jobDetails.getJobIndex(), entry.getJobIndex());
```
dbwiddis, line 183: (Readability) It's not clear what expected/actual are here since the `jobDetails` itself is "actual" up at linen 169.  Can the expected here just be those same four strings?

src/test/java/org/opensearch/jobscheduler/utils/JobDetailsServiceIT.java
```diff
@@ -137,4 +142,57 @@ public void testDeleteNonExistingJobDetails() throws ExecutionException, Interru
...
+                        assertEquals(jobDetails.getJobRunnerAction(), entry.getJobRunnerAction());
+
+                        inProgressFuture.complete(true);
+
+                    }, exception -> { fail(exception.getMessage()); })
+                );
+            }, exception -> { fail(exception.getMessage()); })
+        );
+
+        inProgressFuture.get(JobDetailsService.TIME_OUT_FOR_REQUEST, TimeUnit.SECONDS);
```
dbwiddis, line 195: (Readability/Consistency) If we time out I assume the test fails with an exception.  That's fine and expected, but seems to be different than line 190 and 192 where you explicitly `fail()` on exceptions.   Would be more readable with consistent treatment of these.

Also consider many developers look at the tests to see how they should implement, so it should likely match a "real" execution workflow.

Suggest lines 190 and 192 be `completeExceptionally()` and you have a single try/catch around the get with a `fail()` in the catch block.

src/main/java/org/opensearch/jobscheduler/JobSchedulerPlugin.java
```diff
@@ -151,6 +152,10 @@ public List<ExecutorBuilder<?>> getExecutorBuilders(Settings settings) {
 
     @Override
     public void onIndexModule(IndexModule indexModule) {
+        if (indexModule.getIndex().getName().equals(JobDetailsService.JOB_DETAILS_INDEX_NAME)) {
+            indexModule.addIndexOperationListener(this.jobDetailsService);
+            log.info("JobDetailsService started listening to operations on index {}", indexModule.getIndex().getName());
```
dbwiddis, line 157: This works but since we already checked equality wouldn't it be simpler to use `JobDetailsService.JOB_DETAILS_INDEX_NAME` here?

src/main/java/org/opensearch/jobscheduler/utils/JobDetailsService.java
```diff
@@ -33,31 +33,89 @@
...
+    }
+
+    private void updateIndicesToListen(String jobIndex) {
+        this.indicesToListen.add(jobIndex);
+    }
+
+    /**
+     * Registers or updates a jobDetails entry
+     *
+     * @param extensionId the unique Id for the job details
```
dbwiddis, line 82: Would prefer to use the parameter name `uniqueId` to be consistent (and also in the tests).

src/main/java/org/opensearch/jobscheduler/utils/JobDetailsService.java
```diff
@@ -33,31 +33,89 @@
...
+     * @param jobDetails the jobDetails to register
+     */
+    void updateIndexToJobDetails(String extensionId, JobDetails jobDetails) {
+        if (indexToJobDetails.containsKey(extensionId)) {
```
dbwiddis, line 86: This whole set of conditionals is confusing.  It may be more readable by making the `getJobType()` null check the outer conditional, and using `getOrDefault()` or other map methods.  In any case, the javadoc should be more descriptive of what's happening in here "updates the map to include job details if X, or adds job details if Y"

src/main/java/org/opensearch/jobscheduler/utils/JobDetailsService.java
```diff
@@ -33,31 +33,89 @@
...
+        // Determine if index operation was successful
+        if (result.getResultType().equals(Engine.Result.Type.FAILURE)) {
+            logger.info("Job Details Registration failed for extension {} on index {}", index.id(), shardId.getIndexName());
+            return;
+        }
+
+        // Generate parser using bytesRef from index
+        try {
+            XContentParser parser = XContentType.JSON.xContent()
+                .createParser(NamedXContentRegistry.EMPTY, LoggingDeprecationHandler.INSTANCE, index.source().utf8ToString());
```
dbwiddis, line 111: Do we really not need any XContent from anywhere in this plugin or OpenSearch?  OK if so, just had to ask.

dbwiddis (commented):
LGTM (mostly).  Several readability suggestions.

src/main/java/org/opensearch/jobscheduler/utils/JobDetailsService.java
joshpalis: No the XContentRegistry is not needed here to read the `JobDetails` object from the index. The only NamedXcontentRegistry entry that Job Scheduler registers is the parser for `Schedule` [here](https://github.com/opensearch-project/job-scheduler/blob/6f93b49c88df3476a786216e190702d853bbe69f/src/main/java/org/opensearch/jobscheduler/JobSchedulerPlugin.java#LL179C2-L179C2), which is not used within the `JobDetailsService`

src/main/java/org/opensearch/jobscheduler/utils/JobDetailsService.java
joshpalis: Sure Ill make this change

vibrantvarun (approved):

dbwiddis (dismissed):
LGTM!

src/main/java/org/opensearch/jobscheduler/utils/JobDetailsService.java
```diff
@@ -33,31 +33,87 @@
...
+    public static final String JOB_DETAILS_INDEX_NAME = ".opensearch-plugins-job-details";
     private static final String PLUGINS_JOB_DETAILS_MAPPING_FILE = "/mappings/opensearch_plugins_job_details.json";
 
     public static Long TIME_OUT_FOR_REQUEST = 10L;
     private final Client client;
     private final ClusterService clusterService;
+    private Set<String> indicesToListen;
 
-    public JobDetailsService(final Client client, final ClusterService clusterService) {
+    private static final ConcurrentMap<String, JobDetails> indexToJobDetails = IndexToJobDetails.getInstance();
```
saratvemulapalli, line 59: Trying to understand why does this have to be static? 
Can we ingest this as a component to createComponent and get it from guice ?

src/main/java/org/opensearch/jobscheduler/utils/JobDetailsService.java
```diff
@@ -33,31 +33,87 @@
...
+     * Adds a new JobDetails entry to the indexToJobDetails if jobType is null, or updates an existing JobDetails entry with the jobType
+     *
+     * @param extensionUniqueId the unique Id for the job details
+     * @param jobDetails the jobDetails to register
+     */
+    void updateIndexToJobDetails(String extensionUniqueId, JobDetails jobDetails) {
+        if (jobDetails.getJobType() != null && indexToJobDetails.containsKey(extensionUniqueId)) {
+            // Update existing JobDetails entry with job type
+            JobDetails existingJobDetails = indexToJobDetails.get(extensionUniqueId);
+            existingJobDetails.setJobType(jobDetails.getJobType());
```
saratvemulapalli, line 89: This means an extension could only register 1 job in a life time? 
Wouldn't that break when an extension wants to register multiple jobs?

joshpalis: No, an extension will be able to register multiple jobs. The `indexToJobDetails` contains the initial configuration information for an extension's job to prepare the `JobSweeper` to begin listening to operations on the registered job index, such as the Job Index Name, Job Type, Job Runner Action name and Job Parameter Action name. This information does not change over time. Jobs are registered when extensions index a document to their corresponding job index.

src/main/java/org/opensearch/jobscheduler/transport/GetJobIndexRequest.java
```diff
@@ -29,11 +29,11 @@ public class GetJobIndexRequest extends ActionRequest {
 
     private static String jobRunnerAction;
 
-    private static String extensionId;
+    private static String extensionUniqueId;
 
     public static final String JOB_INDEX = "job_index";
```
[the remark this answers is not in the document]
joshpalis: This was our initial plan as well, however, we have discussed this with @owaiskazi19 and from this discussion we have decided to keep this separate. Please refer to this thread [here](https://github.com/opensearch-project/anomaly-detection/pull/761#discussion_r1050253390) for the rational for this design decision

src/main/java/org/opensearch/jobscheduler/utils/JobDetailsService.java
```diff
@@ -33,31 +33,87 @@
...
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 
-public class JobDetailsService {
+public class JobDetailsService implements IndexingOperationListener {
 
     private static final Logger logger = LogManager.getLogger(JobDetailsService.class);
-    private static final String JOB_DETAILS_INDEX_NAME = ".opensearch-plugins-job-details";
+    public static final String JOB_DETAILS_INDEX_NAME = ".opensearch-plugins-job-details";
```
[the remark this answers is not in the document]
joshpalis: This is a good idea, I'll go ahead and change the index name

src/main/java/org/opensearch/jobscheduler/rest/RestGetJobTypeAction.java
```diff
@@ -73,16 +73,16 @@ protected RestChannelConsumer prepareRequest(RestRequest restRequest, NodeClient
         final JobDetails[] jobDetailsResponseHolder = new JobDetails[1];
 
         String jobType = getJobTypeRequest.getJobType();
-        String extensionId = getJobTypeRequest.getExtensionId();
+        String extensionUniqueId = getJobTypeRequest.getExtensionUniqueId();
```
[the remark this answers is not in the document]
joshpalis: Please refer to this discussion : https://github.com/opensearch-project/anomaly-detection/pull/761#discussion_r1050226041

src/main/java/org/opensearch/jobscheduler/sweeper/JobSweeper.java
```diff
@@ -103,7 +105,8 @@ public JobSweeper(
         NamedXContentRegistry registry,
         Map<String, ScheduledJobProvider> indexToProviders,
         JobScheduler scheduler,
-        LockService lockService
+        LockService lockService,
+        JobDetailsService jobDetailsService
```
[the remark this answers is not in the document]
joshpalis: This is preparational work, we will eventually need `jobDetailsService`  within the `JobSweeper` to configure an `indexToJobProviders` entry with the information from the `indexToJobDetails`

src/main/java/org/opensearch/jobscheduler/JobSchedulerPlugin.java
```diff
@@ -105,7 +105,8 @@ public Collection<Object> createComponents(
             threadPool,
             xContentRegistry,
             this.scheduler,
-            this.lockService
+            this.lockService,
+            this.jobDetailsService
```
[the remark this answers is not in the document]
joshpalis: `JobDetailsService` needs access to the node `client` to make requests and the `ClusterService` for the index routing table. These are both pulled from `createComponents`.

src/test/java/org/opensearch/jobscheduler/TestHelpers.java
```diff
@@ -28,8 +28,7 @@
 
 public class TestHelpers {
 
-    public static final String GET_JOB_INDEX_BASE_DETECTORS_URI = "/_plugins/_job_scheduler/_get/_job_index";
-    public static final String GET_JOB_TYPE_BASE_DETECTORS_URI = "/_plugins/_job_scheduler/_get/_job_type";
+    public static final String GET_JOB_DETAILS_BASE_DETECTORS_URI = "/_plugins/_job_scheduler/_get/_job_details";
```
saratvemulapalli, line 31: Why detectors ;)

saratvemulapalli (approved):
Thanks @joshpalis for the changes.

src/test/java/org/opensearch/jobscheduler/utils/JobDetailsServiceIT.java
```diff
@@ -16,102 +18,115 @@
...
+    private String expectedDocumentId;
+    private String updatedJobIndex;
 
     @Before
     public void setup() {
         this.clusterService = Mockito.mock(ClusterService.class, Mockito.RETURNS_DEEP_STUBS);
-        Mockito.when(this.clusterService.state().routingTable().hasIndex("opensearch-plugins-job-details"))
+        Mockito.when(this.clusterService.state().routingTable().hasIndex(JobDetailsService.JOB_DETAILS_INDEX_NAME))
             .thenReturn(false)
             .thenReturn(true);
```
dbwiddis, line 43: I'm just now coming up to speed on Mockito, but I'm confused about these consecutive `thenReturn()` statements.  This doesn't look right to me but it might be. Is the second one kind of like "else"?

saratvemulapalli (approved):

src/test/java/org/opensearch/jobscheduler/utils/JobDetailsServiceIT.java
joshpalis: Each `thenReturn` corresponds to a single invocation of `clusterService.state().routingTable().hasIndex()`. The first invocation of this method will return `false`, and the second invocation will return `true`

joshpalis:
As per several offline discussions, there have been a few major changes to the job details registration workflow. 

**Changes to REST APIs**
The SPI design allowed Job Scheduler to enforce each extending plugin to submit only one `ScheduledJobProvider` which includes the job index name and job type. Now that we're retrieving these values from extensions via rest apis, it is difficult to enforce extensions to invoke both the `getJobIndex` and `getJobType` sequentially, as Job Details are not indexed to the metadata index until both APIs are invoked. The discussion to keep these APIs separate can be found [here](https://github.com/opensearch-project/anomaly-detection/pull/761#discussion_r1050226041), and while mirroring the plugin architecture's existing extension points may make the transition of plugins to extensions easier, separating these API calls have lead to many drawbacks and an overcomplication of the Job Details registration. As a result, we have removed the `getJobType` API and combined both the `getJobIndex` and `getJobType` into one request.

**Changes to support extensions registering multiple job types**
Previously, the metadata index in which we store job details was keyed with the extension unique ID from which the entry came from. This initial design followed the existing job registration process for plugins. In the case of extensions, there is a possibility that an extension may want to register more than one type of job, and thus in order to support an extension to register multiple times, it was necessary to forgo using the extension ID as the key. As a result, the key shall now be a randomly generated document ID which will be sent back to the caller as a rest response, in order to enable these extensions to further update their Job Details as needed.

CC : @owaiskazi19 @saratvemulapalli @vibrantvarun

saratvemulapalli (approved):

dbwiddis (approved):

PR #306, opened 2023-01-24
[Extensions] Create a proxy ScheduledJobRunner, ScheduledJobParser to invoke corresponding registered Job Detail actions

joshpalis:
### Description
Updates the `indexToJobProviders` with a proxy `ScheduledJobProvider` entry for an extension. Whenever a `JobDetails` entry is indexed, the entry is pulled and registered within an `indexToJobDetails`, updates the `indicesToListen` set with the registered job index, creates a Proxy `ScheduledJobRunner` to invoke the extension's `runJob` implementation, and a Proxy `ScheduledJobParser` to invoke the extension's `parse` implementation.

These actions are invoked via the `ExtensionProxyAction`, which is registered by the `ExtensionsManager` during bootstrap and delegates these requests to the originating extension. 
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-sdk-java/issues/309
 
### Check List
  - [x] All tests pass

### DIFF
--- a/spi/src/main/java/org/opensearch/jobscheduler/spi/JobDocVersion.java
+++ b/spi/src/main/java/org/opensearch/jobscheduler/spi/JobDocVersion.java
@@ -8,12 +8,17 @@
  */
 package org.opensearch.jobscheduler.spi;
 
+import java.io.IOException;
 import java.util.Locale;
+
+import org.opensearch.common.io.stream.StreamInput;
+import org.opensearch.common.io.stream.StreamOutput;
+import org.opensearch.common.io.stream.Writeable;
 
 /**
  * Structure to represent scheduled job document version. JobScheduler use this to determine this job
  */
-public class JobDocVersion implements Comparable<JobDocVersion> {
+public class JobDocVersion implements Comparable<JobDocVersion>, Writeable {
     private final long primaryTerm;
     private final long seqNo;
     private final long version;
@@ -22,6 +27,19 @@ public class JobDocVersion implements Comparable<JobDocVersion> {
         this.primaryTerm = primaryTerm;
         this.seqNo = seqNo;
         this.version = version;
+    }
+
+    public JobDocVersion(StreamInput in) throws IOException {
+        this.primaryTerm = in.readLong();
+        this.seqNo = in.readLong();
+        this.version = in.readLong();
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        out.writeLong(this.primaryTerm);
+        out.writeLong(this.seqNo);
+        out.writeLong(this.version);
     }
 
     public long getPrimaryTerm() {
--- a/spi/src/main/java/org/opensearch/jobscheduler/spi/JobExecutionContext.java
+++ b/spi/src/main/java/org/opensearch/jobscheduler/spi/JobExecutionContext.java
@@ -8,11 +8,15 @@
  */
 package org.opensearch.jobscheduler.spi;
 
+import org.opensearch.common.io.stream.StreamInput;
+import org.opensearch.common.io.stream.StreamOutput;
+import org.opensearch.common.io.stream.Writeable;
 import org.opensearch.jobscheduler.spi.utils.LockService;
 
+import java.io.IOException;
 import java.time.Instant;
 
-public class JobExecutionContext {
+public class JobExecutionContext implements Writeable {
     private final Instant expectedExecutionTime;
     private final JobDocVersion jobVersion;
     private final LockService lockService;
@@ -31,6 +35,22 @@ public class JobExecutionContext {
         this.lockService = lockService;
         this.jobIndexName = jobIndexName;
         this.jobId = jobId;
+    }
+
+    public JobExecutionContext(StreamInput in) throws IOException {
+        this.expectedExecutionTime = in.readInstant();
+        this.jobVersion = new JobDocVersion(in);
+        this.lockService = null;
+        this.jobIndexName = in.readString();
+        this.jobId = in.readString();
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        out.writeInstant(this.expectedExecutionTime);
+        this.jobVersion.writeTo(out);
+        out.writeString(this.jobIndexName);
+        out.writeString(this.jobId);
     }
 
     public Instant getExpectedExecutionTime() {
@@ -52,4 +72,5 @@ public class JobExecutionContext {
     public String getJobId() {
         return this.jobId;
     }
+
 }
--- a/src/main/java/org/opensearch/jobscheduler/model/ExtensionJobParameter.java
+++ b/src/main/java/org/opensearch/jobscheduler/model/ExtensionJobParameter.java
@@ -1,0 +1,124 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.jobscheduler.model;
+
+import java.io.IOException;
+import java.time.Instant;
+
+import org.opensearch.common.io.stream.StreamInput;
+import org.opensearch.common.io.stream.StreamOutput;
+import org.opensearch.common.io.stream.Writeable;
+import org.opensearch.common.xcontent.XContentBuilder;
+import org.opensearch.jobscheduler.spi.ScheduledJobParameter;
+import org.opensearch.jobscheduler.spi.schedule.CronSchedule;
+import org.opensearch.jobscheduler.spi.schedule.IntervalSchedule;
+import org.opensearch.jobscheduler.spi.schedule.Schedule;
+import org.opensearch.jobscheduler.spi.ScheduledJobParameter;
+
+public class ExtensionJobParameter implements ScheduledJobParameter, Writeable {
+
+    enum ScheduleType {
+        CRON,
+        INTERVAL
+    }
+
+    private String extensionJobName;
+
+    private Schedule extensionJobSchedule;
+
+    private Instant lastUpdateTime;
+
+    private Instant enabledTime;
+
+    private boolean isEnabled;
+
+    private Long lockDurationSeconds;
+
+    private Double jitter;
+
+    public ExtensionJobParameter(
+        String extensionJobName,
+        Schedule extensionJobSchedule,
+        Instant lastUpdateTime,
+        Instant enabledTime,
+        boolean isEnabled,
+        Long lockDurationSeconds,
+        Double jitter
+    ) {
+        this.extensionJobName = extensionJobName;
+        this.extensionJobSchedule = extensionJobSchedule;
+        this.lastUpdateTime = lastUpdateTime;
+        this.enabledTime = enabledTime;
+        this.isEnabled = isEnabled;
+        this.lockDurationSeconds = lockDurationSeconds;
+        this.jitter = jitter;
+    }
+
+    public ExtensionJobParameter(StreamInput in) throws IOException {
+        this.extensionJobName = in.readString();
+        if (in.readEnum(ExtensionJobParameter.ScheduleType.class) == ScheduleType.CRON) {
+            this.extensionJobSchedule = new CronSchedule(in);
+        } else {
+            this.extensionJobSchedule = new IntervalSchedule(in);
+        }
+        this.lastUpdateTime = in.readInstant();
+        this.enabledTime = in.readInstant();
+        this.isEnabled = in.readBoolean();
+        this.lockDurationSeconds = in.readLong();
+        this.jitter = in.readDouble();
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        out.writeString(this.extensionJobName);
+        if (this.extensionJobSchedule instanceof CronSchedule) {
+            out.writeEnum(ScheduleType.CRON);
+        } else {
+            out.writeEnum(ScheduleType.INTERVAL);
+        }
+        this.extensionJobSchedule.writeTo(out);
+        out.writeInstant(this.lastUpdateTime);
+        out.writeInstant(this.enabledTime);
+        out.writeBoolean(this.isEnabled);
+        out.writeLong(this.lockDurationSeconds);
+        out.writeDouble(this.jitter);
+    }
+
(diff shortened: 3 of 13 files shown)


src/main/java/org/opensearch/jobscheduler/model/ExtensionJobParameter.java
```diff
@@ -0,0 +1,141 @@
...
+
+    private Schedule extensionJobSchedule;
+
+    private Instant lastUpdateTime;
+
+    private Instant enabledTime;
+
+    private boolean isEnabled;
+
+    private Long lockDurationSeconds;
```
dbwiddis, line 45: Why is this `Long` and not `long`?  Introduces the possibility of it being `null` (vs. default 0) and would require special handling in the `Writable` implementation (e.g., `writeOptionalLong`) if null is a desired/possible value.  (In which case `@Nullable` annotation would be helpful.)

src/main/java/org/opensearch/jobscheduler/model/ExtensionJobParameter.java
```diff
@@ -0,0 +1,141 @@
...
+
+    private Double jitter;
```
dbwiddis, line 47: Similar question as above about why the boxed type.

src/main/java/org/opensearch/jobscheduler/model/ExtensionJobParameter.java
```diff
@@ -0,0 +1,141 @@
...
+        out.writeInstant(this.enabledTime);
+        out.writeBoolean(this.isEnabled);
+        out.writeLong(this.lockDurationSeconds);
+        out.writeDouble(this.jitter);
+    }
+
+    @Override
+    public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
+        // no op
+        return null;
```
dbwiddis, line 113: This seems dangerous.  Is there a better "empty builder" constant we can return here?

src/main/java/org/opensearch/jobscheduler/transport/ExtensionJobActionRequest.java
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * Takes in an object of type T that extends {@link Writeable} and converts the request to a byte array
+     *
+     * @param <T> a class that extends writeable
+     * @param actionParams the action parameters to be serialized
+     * @throws IOException if serialization fails
+     * @return the byte array of the parameters
+     */
+    public static <T extends Writeable> byte[] convertParamsToBytes(T actionParams) throws IOException {
```
dbwiddis, line 43: No issues with this method to produce a byte stream, but since this is effectively the same as creating a `Writeable` object, why not just pass the `actionParams` directly as a writeable object?   The only benefit I see to this is that it allows for a more flexible signature (a byte array) but that loses the type safety of the individual elements of it, and introduces complications handling the different array lengths when it's deserialized.

src/main/java/org/opensearch/jobscheduler/transport/ExtensionJobActionResponse.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Takes in an object of type T that extends {@link Writeable} and converts the response to a byte array
+     *
+     * @param <T> a class that extends writeable
+     * @param actionResponse the action parameters to be serialized
+     * @throws IOException if serialization fails
+     * @return the byte array of the parameters
+     */
+    public static <T extends Writeable> byte[] convertResponseToBytes(T actionResponse) throws IOException {
```
dbwiddis, line 42: this duplicates the static method in `ExtensionJobActionRequest`.  Seems it could just be a static method on the `Writeable` interface itself if it's that useful.  Or at least at some common parent of `ExtensionActionRequest` and `ExtensionActionResponse`.

I'm still not convinced it has any benefit beyond the `Writeable` interface that lets it be deserialized somewhere. Where is this byte array handled?

src/main/java/org/opensearch/jobscheduler/utils/JobDetailsService.java
```diff
@@ -86,6 +132,129 @@ void updateIndexToJobDetails(String documentId, JobDetails jobDetails) {
...
+
+                    })
+                );
+
+                // Stall execution until request completes or times out
+                try {
+                    inProgressFuture.orTimeout(JobDetailsService.TIME_OUT_FOR_REQUEST, TimeUnit.SECONDS).join();
+                } catch (CompletionException e) {
+                    if (e.getCause() instanceof TimeoutException) {
+                        logger.info("Request timed out with an exception ", e);
```
dbwiddis, line 184: the way this is handled, this will just log this and not throw an exception.  You may not want "else" below this.

src/main/java/org/opensearch/jobscheduler/transport/JobParameterRequest.java
```diff
@@ -0,0 +1,112 @@
...
+     * @param jobDocVersion the job document version
+     * @throws IOException IOException when message de-serialization fails.
+     */
+    public JobParameterRequest(String accessToken, XContentParser jobParser, String id, JobDocVersion jobDocVersion) throws IOException {
+
+        // Extract jobSource bytesRef from xContentParser
+        XContentBuilder builder = XContentFactory.jsonBuilder();
+        builder.copyCurrentStructure(jobParser);
+        this.jobSource = BytesReference.bytes(builder);
+        this.accessToken = accessToken;
```
saratvemulapalli, line 61: I would leave the details of `accessToken` to @dbwiddis. 
My preference is to not care about security at this point and get the communication working and we can add security later. There might be lot more implications we have to consider like:
- How are we gating JS Rest APIs
- Who can modify a job parameter, job index (a.k.a any meta data) etc

spi/src/main/java/org/opensearch/jobscheduler/spi/JobDocVersion.java
```diff
@@ -24,6 +29,19 @@ public JobDocVersion(long primaryTerm, long seqNo, long version) {
         this.version = version;
     }
 
+    public JobDocVersion(StreamInput in) throws IOException {
```
saratvemulapalli, line 32: What is JobDocVersion, who uses this internally? 
Why should an extension care about this internal implementation of JS?

joshpalis: The `JobDocVersion` represents the version of a particular job entry, which JobScheduler uses to determine whether or not a particular job entry is outdated, this can happen when a job configuration is updated by the plugin. The `JobSweeper` maintains a map of jobs that it has "swept" (schedule/deschedule) which occurs every time a plugin indexes/updates another job to it's registered job index. 

The reason why the `JobDocVersion` is included as part of the request to extensions is because it is one of the parameters of `ScheduledJobParser.parse(XcontentParser, docId, jobDocVersion)`.

While the Anomaly Detection implementation of the `ScheduledJobParser` utilizes only the provided `xContentParser`, I felt that it would be necessary to add support to transport both the `docId` and also the `JobDocVersion` as well, in case an extension required this information.

joshpalis: For instance, ISM requires the `jobDocVersion` [here](https://github.com/opensearch-project/index-management/blob/76d530fe1c4a46d0dab8a179557f30b31bcc2e91/src/main/kotlin/org/opensearch/indexmanagement/IndexManagementPlugin.kt#LL237C46-L237C60) within their `ScheduledJobParser` implementation

src/main/java/org/opensearch/jobscheduler/model/ExtensionJobParameter.java
joshpalis: meant to use `long`, thanks for catching this, I'll go ahead and make this change

joshpalis: The `ExtensionJobParameter` implements the `ScheduledJobParameter` interface. Within this interface, the `lockDurationSeconds` and `jitter` values are nullable, in the case that a scheduled job does not require a lock, or if a job execution delay isn't necessary. 

Thank you for pointing this out though, I'll go ahead and modify the writeTo method so that it writes an optional long and double as well

joshpalis:
```
/home/runner/work/job-scheduler/job-scheduler/src/test/java/org/opensearch/jobscheduler/utils/JobDetailsServiceIT.java:308: error: ExtensionActionRequest(StreamInput) is not public in ExtensionActionRequest; cannot be accessed from outside package

                actionRequest = new ExtensionActionRequest(in);
> Task :compileTestJava FAILED
                                ^
/home/runner/work/job-scheduler/job-scheduler/src/test/java/org/opensearch/jobscheduler/utils/JobDetailsServiceIT.java:343: error: ExtensionActionRequest(StreamInput) is not public in ExtensionActionRequest; cannot be accessed from outside package
                actionRequest = new ExtensionActionRequest(in);
```
This particular constructor has been made public [here](https://github.com/opensearch-project/OpenSearch/pull/6094). Seems that checks are failing due to stale OS 3.0.0 artifacts, will re-run all failed jobs once a successful jenkins build has been made.

spi/src/main/java/org/opensearch/jobscheduler/spi/JobExecutionContext.java
```diff
@@ -8,11 +8,15 @@
...
 import org.opensearch.jobscheduler.spi.utils.LockService;
 
+import java.io.IOException;
 import java.time.Instant;
 
-public class JobExecutionContext {
+public class JobExecutionContext implements Writeable {
     private final Instant expectedExecutionTime;
     private final JobDocVersion jobVersion;
     private final LockService lockService;
```
dbwiddis, line 22: For my own education, what is this `LockService` used for? It is not included in the `Writeable` implementation so it is not sent over transport.  It is only set (final) in the constructor and read via a getter (and could be null if it's been sent over transport).

src/main/java/org/opensearch/jobscheduler/model/ExtensionJobParameter.java
```diff
@@ -0,0 +1,160 @@
...
+import org.opensearch.jobscheduler.spi.schedule.Schedule;
+import org.opensearch.jobscheduler.spi.ScheduledJobParameter;
+
+/**
+ * A {@link Writeable} ScheduledJobParameter used to transport job parameters between OpenSearch and Extensions
+ *
+ */
+public class ExtensionJobParameter implements ScheduledJobParameter, Writeable {
+
+    enum ScheduleType {
```
dbwiddis, line 31: Is package private access intentional here?

src/main/java/org/opensearch/jobscheduler/model/ExtensionJobParameter.java
```diff
@@ -0,0 +1,160 @@
...
+        CRON,
```
dbwiddis, line 32: (Docs) Assuming this is API and not package private, would be helpful if these enum values had a javadoc on them describing their use, e.g., CRON is at a scheduled time of day, INTERVAL is repeating every N TimeUnits.

I assume this enum is copied from an existing implementation and these are the only two options?

src/main/java/org/opensearch/jobscheduler/model/ExtensionJobParameter.java
```diff
@@ -0,0 +1,160 @@
...
+    @Override
+    public Schedule getSchedule() {
+        return this.schedule;
+    }
+
+    @Override
+    public boolean isEnabled() {
+        return this.isEnabled;
+    }
+
```
dbwiddis, line 159: Do you want getters for the durationSeconds and jitter fields?   (Could document nullable or make Optionals.)

src/main/java/org/opensearch/jobscheduler/transport/JobParameterRequest.java
dbwiddis: Agree the "details" will be implemented later but I think it's safe to put a placeholder String here for now as part of the "communication".  We're doing the same thing with the Rest Handler communication with a placeholder string and eventual authentication integration based on it.

src/main/java/org/opensearch/jobscheduler/transport/JobParameterResponse.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    /**
+     * Instantiates a new Job Parameter Response by wrapping the given byte array within a {@link StreamInput}
+     *
+     * @param responseParams in bytes array used to de-serialize the message.
+     * @throws IOException when message de-serialization fails.
+     */
+    public JobParameterResponse(byte[] responseParams) throws IOException {
+        this(StreamInput.wrap(responseParams));
```
dbwiddis, line 54: This is a cool implementation I haven't seen before!

dbwiddis (dismissed):
"Approved with suggestions": 

LGTM, a few questions and suggestions you may want to consider but I wouldn't hold up merging for them.

spi/src/main/java/org/opensearch/jobscheduler/spi/JobExecutionContext.java
joshpalis: Yes, it is not included in the `Writeable` implementation for a couple of reasons, most of which have been documented in this issue here : https://github.com/opensearch-project/opensearch-sdk-java/issues/363. 

When a plugin is prompted to run their job, this `JobExecutionContext` is provided which includes information such as the job version, id, expected execution time, and also the `LockService` object. This is used by the plugin to invoke `aquireLock` prior to executing their job (which retrieves a lock model object from the lock service metadata index) or `releaseLock` after their job execution has been completed (which updates the lock index entry so that another job can be enabled to run). 

For extensions, we still want to provide the  `JobExecutionContext` when triggering an extension's `ScheduledJobRunner.runJob` implementation, hence why this class was modified to implement `Writeable`, however we cannot send the `LockService` over transport. To enable extensions to still acquire/release this lock, we shall expose two REST APIs on Job Scheduler to achieve the same, which is tracked as work items 5 and 6 within our meta issue for the communication mechanism.

src/main/java/org/opensearch/jobscheduler/model/ExtensionJobParameter.java
joshpalis: Great catch, unfortunately I missed adding those, I'll go ahead and do so.

dbwiddis (dismissed):

dbwiddis (approved):

saratvemulapalli (approved):

PR #311, opened 2023-02-07
[Extensions] Exposes a GetLock REST API to enable extensions to acquire a lock model for their job execution

joshpalis:
### Description
Exposes a GetLock Rest API to enable extensions to acquire a lock model for their job execution
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-sdk-java/issues/363
 
### Check List
  - [X] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/jobscheduler/rest/RestGetLockAction.java
+++ b/src/main/java/org/opensearch/jobscheduler/rest/RestGetLockAction.java
@@ -37,16 +37,18 @@ import com.google.common.collect.ImmutableList;
 import static org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
 import static org.opensearch.rest.RestRequest.Method.GET;
 
+import static org.opensearch.jobscheduler.spi.LockModel.GET_LOCK_ACTION;
+import static org.opensearch.jobscheduler.spi.LockModel.SEQUENCE_NUMBER;
+import static org.opensearch.jobscheduler.spi.LockModel.SEQUENCE_NUMBER;
+import static org.opensearch.jobscheduler.spi.LockModel.PRIMARY_TERM;
+import static org.opensearch.jobscheduler.spi.LockModel.LOCK_ID;
+import static org.opensearch.jobscheduler.spi.LockModel.LOCK_MODEL;
+
 /**
  * This class consists of the REST handler to GET a lock model for extensions
  */
 public class RestGetLockAction extends BaseRestHandler {
 
-    public static final String GET_LOCK_ACTION = "get_lock_action";
-    public static final String SEQUENCE_NUMBER = "seq_no";
-    public static final String PRIMARY_TERM = "primary_term";
-    public static final String LOCK_ID = "lock_id";
-    public static final String LOCK_MODEL = "lock_model";
     public static LockModel lockModelResponseHolder;
 
     private final Logger logger = LogManager.getLogger(RestGetLockAction.class);
@@ -109,12 +111,12 @@ public class RestGetLockAction extends BaseRestHandler {
         }
 
         return channel -> {
-            // Prepare response
-            XContentBuilder builder = channel.newBuilder();
-            RestStatus restStatus = RestStatus.OK;
-            String restResponseString = lockModelResponseHolder != null ? "success" : "failed";
             BytesRestResponse bytesRestResponse;
-            try {
+            try (XContentBuilder builder = channel.newBuilder()) {
+                // Prepare response
+                RestStatus restStatus = RestStatus.OK;
+                String restResponseString = lockModelResponseHolder != null ? "success" : "failed";
+
                 builder.startObject();
                 builder.field("response", restResponseString);
                 if (restResponseString.equals("success")) {
@@ -133,10 +135,9 @@ public class RestGetLockAction extends BaseRestHandler {
                 }
                 builder.endObject();
                 bytesRestResponse = new BytesRestResponse(restStatus, builder);
-            } finally {
-                builder.close();
+                channel.sendResponse(bytesRestResponse);
             }
-            channel.sendResponse(bytesRestResponse);
+
         };
     }
 }
--- a/src/main/java/org/opensearch/jobscheduler/rest/RestGetLockAction.java
+++ b/src/main/java/org/opensearch/jobscheduler/rest/RestGetLockAction.java
@@ -47,6 +47,7 @@ public class RestGetLockAction extends BaseRestHandler {
     public static final String PRIMARY_TERM = "primary_term";
     public static final String LOCK_ID = "lock_id";
     public static final String LOCK_MODEL = "lock_model";
+    public static LockModel lockModelResponseHolder;
 
     private final Logger logger = LogManager.getLogger(RestGetLockAction.class);
 
@@ -63,7 +64,7 @@ public class RestGetLockAction extends BaseRestHandler {
 
     @Override
     public List<Route> routes() {
-        return ImmutableList.of(new Route(GET, String.format(Locale.ROOT, "%s/%s", JobSchedulerPlugin.JS_BASE_URI, "_get/_lock")));
+        return ImmutableList.of(new Route(GET, String.format(Locale.ROOT, "%s/%s", JobSchedulerPlugin.JS_BASE_URI, "_lock")));
     }
 
     @Override
@@ -78,14 +79,13 @@ public class RestGetLockAction extends BaseRestHandler {
         long lockDurationSeconds = acquireLockRequest.getLockDurationSeconds();
 
         // Process acquire lock request
-        final LockModel[] lockModelResponseHolder = new LockModel[1];
-        CompletableFuture<LockModel[]> inProgressFuture = new CompletableFuture<>();
+        CompletableFuture<LockModel> inProgressFuture = new CompletableFuture<>();
         lockService.acquireLockWithId(jobIndexName, lockDurationSeconds, jobIndexName, new ActionListener<>() {
             @Override
             public void onResponse(LockModel lockModel) {
 
                 // set lockModel Response
-                lockModelResponseHolder[0] = lockModel;
+                lockModelResponseHolder = lockModel;
                 inProgressFuture.complete(lockModelResponseHolder);
             }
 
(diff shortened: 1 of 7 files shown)


src/test/java/org/opensearch/jobscheduler/rest/RestGetLockActionIT.java
```diff
@@ -0,0 +1,115 @@
...
+        super.setUp();
+        this.clusterService = Mockito.mock(ClusterService.class, Mockito.RETURNS_DEEP_STUBS);
+        Mockito.when(this.clusterService.state().routingTable().hasIndex(".opendistro-job-scheduler-lock")).thenReturn(true);
+        this.lockService = new LockService(Mockito.mock(NodeClient.class), clusterService);
+        this.getLockAction = new RestGetLockAction(this.lockService);
+        this.getLockPath = String.format(Locale.ROOT, "%s/%s", JobSchedulerPlugin.JS_BASE_URI, "_get/_lock");
+        this.testJobId = "testJobId";
+        this.testJobIndexName = "testJobIndexName";
+        this.testLockDurationSeconds = 1L;
+        this.requestBody = "{\"job_id\":\"test\",\"job_index_name\":\"test\",\"lock_duration_seconds\":\"1\"}";
```
dbwiddis, line 62: Having the same name for different things you're testing doesn't help you deconflict whether you assigned them to the wrong things.  Why not use the `testJobId` and `testJobIndexName` values here?  Not that they are ever used, of course, you could probably put any string here and it would work. :)

dbwiddis (approved):
Approved with a nitpicky but irrelevant suggestion.

src/main/java/org/opensearch/jobscheduler/rest/RestGetLockAction.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return GET_LOCK_ACTION;
+    }
+
+    @Override
+    public List<Route> routes() {
+        return ImmutableList.of(new Route(GET, String.format(Locale.ROOT, "%s/%s", JobSchedulerPlugin.JS_BASE_URI, "_get/_lock")));
```
owaiskazi19, line 66: It's a GET handler. No need to add the keyword in the uri
```suggestion
        return ImmutableList.of(new Route(GET, String.format(Locale.ROOT, "%s/%s", JobSchedulerPlugin.JS_BASE_URI, "_lock")));
```

src/main/java/org/opensearch/jobscheduler/rest/RestGetLockAction.java
```diff
@@ -0,0 +1,142 @@
...
+        ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
+
+        // Deserialize acquire lock request
+        AcquireLockRequest acquireLockRequest = AcquireLockRequest.parse(parser);
+        String jobId = acquireLockRequest.getJobId();
+        String jobIndexName = acquireLockRequest.getJobIndexName();
+        long lockDurationSeconds = acquireLockRequest.getLockDurationSeconds();
+
+        // Process acquire lock request
+        final LockModel[] lockModelResponseHolder = new LockModel[1];
```
owaiskazi19, line 81: Create a global variable `lockModelResponseHolder` instead of a size 1 array.

src/main/java/org/opensearch/jobscheduler/rest/RestGetLockAction.java
```diff
@@ -0,0 +1,142 @@
...
+                logger.info("Could not acquire lock with ID : " + jobId, e);
+                inProgressFuture.completeExceptionally(e);
+            }
+        });
+
+        try {
+            inProgressFuture.orTimeout(JobDetailsService.TIME_OUT_FOR_REQUEST, TimeUnit.SECONDS).join();
+        } catch (CompletionException e) {
+            if (e.getCause() instanceof TimeoutException) {
+                logger.info(" Request timed out with an exception ", e);
```
owaiskazi19, line 103:
```suggestion
                logger.error(" Request timed out with an exception ", e);
```

src/main/java/org/opensearch/jobscheduler/rest/RestGetLockAction.java
```diff
@@ -0,0 +1,142 @@
...
+            } else {
+                throw e;
+            }
+        } catch (Exception e) {
+            logger.info(" Could not process acquire lock request due to exception ", e);
```
owaiskazi19, line 108:
```suggestion
            logger.error(" Could not process acquire lock request due to exception ", e);
```

src/test/java/org/opensearch/jobscheduler/multinode/GetLockMultiNodeRestIT.java
```diff
@@ -0,0 +1,72 @@
...
+
+        // Send initial request to ensure lock index has been created
+        Response response = TestHelpers.makeRequest(
+            client(),
+            "GET",
+            TestHelpers.GET_LOCK_BASE_URI,
+            ImmutableMap.of(),
+            TestHelpers.toHttpEntity(generateRequestBody(initialJobIndexName, intialJobId)),
+            null
+        );
```
owaiskazi19, line 38: Can we move this under `setUp()`?

src/test/java/org/opensearch/jobscheduler/multinode/GetLockMultiNodeRestIT.java
```diff
@@ -0,0 +1,72 @@
...
+
+        String initialLockId = validateResponseAndGetLockId(entityAsMap(response));
+        assertEquals(generateExpectedLockId(initialJobIndexName, intialJobId), initialLockId);
+
+        // Submit 100 requests to generate new lock models for different job indexes
+        for (int i = 0; i < 100; i++) {
```
owaiskazi19, line 44: Let's keep the count to 10 to avoid timeout for test

owaiskazi19 (changes requested):

owaiskazi19 (dismissed):

vibrantvarun (approved):
LGTM

src/main/java/org/opensearch/jobscheduler/rest/RestGetLockAction.java
```diff
@@ -0,0 +1,142 @@
...
+            logger.error(" Could not process acquire lock request due to exception ", e);
+        }
+
+        return channel -> {
+            // Prepare response
+            XContentBuilder builder = channel.newBuilder();
+            RestStatus restStatus = RestStatus.OK;
+            String restResponseString = lockModelResponseHolder != null ? "success" : "failed";
+            BytesRestResponse bytesRestResponse;
+            try {
```
dbwiddis, line 117: Can we turn this try... finally into a try with resources?
```java
try (XContentBuilder = channel.newBuilder()) {
   // code here
}
```

src/main/java/org/opensearch/jobscheduler/transport/AcquireLockRequest.java
```diff
@@ -0,0 +1,100 @@
...
+    public static final String LOCK_DURATION_SECONDS = "lock_duration_seconds";
+
+    public AcquireLockRequest(StreamInput in) throws IOException {
+        super(in);
+        this.jobId = in.readString();
+        this.jobIndexName = in.readString();
+        this.lockDurationSeconds = in.readLong();
+    }
+
+    public AcquireLockRequest(String jobId, String jobIndexName, long lockDurationSeconds) {
```
dbwiddis, line 39: (Style) Pure preference here, I'd like this constructor first (near the fields) since they are logical pairs, and the StreamInput constructor right next to the `writeTo()` method since they are logical pairs.

src/main/java/org/opensearch/jobscheduler/transport/AcquireLockRequest.java
```diff
@@ -0,0 +1,100 @@
...
+import java.io.IOException;
+import java.util.Objects;
+import org.opensearch.action.ActionRequest;
+import org.opensearch.action.ActionRequestValidationException;
+import org.opensearch.common.io.stream.StreamInput;
+import org.opensearch.common.io.stream.StreamOutput;
+import org.opensearch.common.xcontent.XContentParser;
+import org.opensearch.common.xcontent.XContentParserUtils;
+
+public class AcquireLockRequest extends ActionRequest {
```
dbwiddis, line 20: We aren't enforcing javadocs here but we need some.  We probably should describe what this class does and in the constructor what each parameter means.

src/test/java/org/opensearch/jobscheduler/rest/RestGetLockActionIT.java
```diff
@@ -0,0 +1,121 @@
...
+        builder.startObject();
+        builder.field(AcquireLockRequest.JOB_ID, acquireLockRequest.getJobId());
+        builder.field(AcquireLockRequest.JOB_INDEX_NAME, acquireLockRequest.getJobIndexName());
+        builder.field(AcquireLockRequest.LOCK_DURATION_SECONDS, acquireLockRequest.getLockDurationSeconds());
+        builder.endObject();
+
+        // Test request serde logic
+        XContentParser parser = XContentType.JSON.xContent()
+            .createParser(NamedXContentRegistry.EMPTY, LoggingDeprecationHandler.INSTANCE, TestHelpers.xContentBuilderToString(builder));
+        parser.nextToken();
```
dbwiddis, line 97: this isn't obvious why it's needed, and in other code I've seen this wrapped in an assertion that it's the start of the object or something like that. Might be useful here.

dbwiddis (approved):
Approved with some suggestions

saratvemulapalli (commented):
I did one pass, overall looks good to me.

src/main/java/org/opensearch/jobscheduler/rest/RestGetLockAction.java
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * This class consists of the REST handler to GET a lock model for extensions
+ */
+public class RestGetLockAction extends BaseRestHandler {
+
+    public static final String GET_LOCK_ACTION = "get_lock_action";
+    public static final String SEQUENCE_NUMBER = "seq_no";
+    public static final String PRIMARY_TERM = "primary_term";
+    public static final String LOCK_ID = "lock_id";
+    public static final String LOCK_MODEL = "lock_model";
```
saratvemulapalli, line 49: nit: Aren't these already been defined by lock service? 
Ideally they should be a property of that class.

src/main/java/org/opensearch/jobscheduler/transport/AcquireLockRequest.java
saratvemulapalli, line 20: Is AcquireLockRequest really being used in transport?

src/main/java/org/opensearch/jobscheduler/rest/RestGetLockAction.java
joshpalis: These are not defined by the lock service, but I shall move these over to the LockModel class

src/main/java/org/opensearch/jobscheduler/transport/AcquireLockRequest.java
joshpalis: @vibrantvarun will be reorganizing these classes in their appropriate directory in his PR after this is merged in

dbwiddis (approved):

saratvemulapalli (approved):

PR #312, opened 2023-02-10
Communication Mechanism Work Item 6 Release lock api

vibrantvarun:
### Description
Release Lock API
 
### Issues Resolved
[https://github.com/opensearch-project/opensearch-sdk-java/issues/370](https://github.com/opensearch-project/opensearch-sdk-java/issues/370)
 
### Check List
  - [x] All tests pass
- [x] New functionality has been documented.
  - [x] New functionality has javadoc added


src/test/java/org/opensearch/jobscheduler/multinode/ReleaseLockActionMultiNodeRestIT.java
```diff
@@ -0,0 +1,60 @@
...
+            TestHelpers.toHttpEntity(TestHelpers.generateAcquireLockRequestBody(initialJobIndexName, initialJobId)),
+            null
+        );
+    }
+
+    public void testReleaseLockRestAPI() throws Exception {
+        String initialLockId = validateResponseAndGetLockId(entityAsMap(this.initialGetLockResponse));
+        assertEquals(TestHelpers.generateExpectedLockId(initialJobIndexName, initialJobId), initialLockId);
+        // Submit 10 requests to release locks for different job indexes
+        Response releaseLockResponse = TestHelpers.makeRequest(
```
joshpalis, line 46: This test only generates one request to release the lock that was created during set up. I suggest modifying the set up to generate 10 request to generate the locks and handle releasing them here in this test

src/test/java/org/opensearch/jobscheduler/TestHelpers.java
```diff
@@ -86,4 +92,16 @@ public static Response makeRequest(
         return client.performRequest(request);
     }
 
+    public static LockModel randomLockModel() {
```
joshpalis, line 95: where is this being used?

joshpalis (commented):
Doing an initial pass

src/test/java/org/opensearch/jobscheduler/multinode/ReleaseLockActionMultiNodeRestIT.java
vibrantvarun: Yes I have to remove that comment. As I have only used acquirelock api once so corresponding to that I have used releaselock api also once. Release lock api can be triggered from any node in Opensearch internally. So it serves the multinode test purpose.

src/test/java/org/opensearch/jobscheduler/TestHelpers.java
vibrantvarun: Yes it not needed therefore removing it

vibrantvarun:
Just a premise, This PR has below changes
1) Reorganizing request files in request package and response files in response package
2) Release Lock API
3) Removing _get from RestGetJobDetailsAction api route as the api is already a GET api so it is not needed to be in path

@joshpalis @dbwiddis @saratvemulapalli

src/main/java/org/opensearch/jobscheduler/rest/action/RestGetJobDetailsAction.java
```diff
@@ -61,39 +63,30 @@ public String getName() {
...
 
     @Override
-    protected RestChannelConsumer prepareRequest(RestRequest restRequest, NodeClient client) throws IOException {
+    public RestChannelConsumer prepareRequest(RestRequest restRequest, NodeClient client) throws IOException {
         XContentParser parser = restRequest.contentParser();
         ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
 
         GetJobDetailsRequest getJobDetailsRequest = GetJobDetailsRequest.parse(parser);
-
-        final String[] jobDetailsResponseHolder = new String[1];
```
dbwiddis, line 87: I'm guessing this is leftover from a "need a mutable string" Countdown Latch implementation?

dbwiddis (changes requested):
LGTM but a few questions for my own awareness, and strong suggestion of improvement.  Feel free to push back if I missed something.

src/main/java/org/opensearch/jobscheduler/rest/action/RestGetJobDetailsAction.java
```diff
@@ -61,39 +63,30 @@ public String getName() {
     public List<Route> routes() {
         return ImmutableList.of(
             // New Job Details Entry Request
-            new Route(PUT, String.format(Locale.ROOT, "%s/%s", JobSchedulerPlugin.JS_BASE_URI, "_get/_job_details")),
+            new Route(PUT, String.format(Locale.ROOT, "%s/%s", JobSchedulerPlugin.JS_BASE_URI, "_job_details")),
```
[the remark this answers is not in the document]
vibrantvarun: Answer 1) /_plugins/_job_scheduler/_get/_job_details now changed to /_plugins/_job_scheduler/_job_details
 It has extra _get in its path. So if the API is already a GET Api then it does not make sense to keep _get path

Answer 2) Yes it was a mistake. 

The API is PUT because it updates job details of extension in job scheduler meta deta index. However, essentially what is does is get job details from extension from extension and then update in its index. That's the reason it is PUT on JS end

src/main/java/org/opensearch/jobscheduler/rest/action/RestGetJobDetailsAction.java
vibrantvarun: No actually as we use completetable future instead of countdown latch now, so when we are setting a response of the async call in the local variable then It says to create a final immutable array and then do it. Instead the ideal way is to create a global variable in the class and achieve it. So that's what I did here

src/main/java/org/opensearch/jobscheduler/rest/action/RestReleaseLockAction.java
```diff
@@ -0,0 +1,140 @@
...
+public class RestReleaseLockAction extends BaseRestHandler {
+
+    public static final String RELEASE_LOCK_ACTION = "release_lock_action";
+    public static final String LOCK_ID = "lock_id";
+
+    private final Logger logger = LogManager.getLogger(RestReleaseLockAction.class);
+
+    private LockService lockService;
+
+    private LockModel releaseLock;
```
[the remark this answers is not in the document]
vibrantvarun: Yeah I'll need to do some testing on my end to make this thing possible. As this change includes 3 api's under radar. So after testing I will raise an another PR.

Thanks for suggesting this

src/main/java/org/opensearch/jobscheduler/rest/action/RestGetLockAction.java
```diff
@@ -82,21 +81,13 @@ protected RestChannelConsumer prepareRequest(RestRequest restRequest, NodeClient
...
-            }
-
-            @Override
-            public void onFailure(Exception e) {
-                logger.info("Could not acquire lock with ID : " + jobId, e);
-                inProgressFuture.completeExceptionally(e);
-            }
-        });
+        lockService.acquireLockWithId(jobIndexName, lockDurationSeconds, jobId, ActionListener.wrap(lockModel -> {
+            lockModelResponseHolder = lockModel;
```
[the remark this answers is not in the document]
vibrantvarun: Yeah I'll need to do some testing on my end to make this thing possible. As this change includes 3 api's under radar. So after testing I will raise an another PR.

Thanks for suggesting this

src/main/java/org/opensearch/jobscheduler/rest/action/RestGetJobDetailsAction.java
dbwiddis: yeah, I was just saying I think we had the array in there earlier, then it stayed an array when we moved to completable futures.  This change is good.

dbwiddis (approved):
Approving based on the intent to refactor the temporary fields to completable future `.get()` in a future PR.

src/main/java/org/opensearch/jobscheduler/rest/action/RestGetLockAction.java
```diff
@@ -70,7 +69,7 @@ public List<Route> routes() {
     }
 
     @Override
-    protected RestChannelConsumer prepareRequest(RestRequest restRequest, NodeClient client) throws IOException {
+    public RestChannelConsumer prepareRequest(RestRequest restRequest, NodeClient client) throws IOException {
```
saratvemulapalli, line 69: Curious why `public`? Does it matter?
This interface is anyway called internally by BaseRestHandler[1].

[1] https://github.com/opensearch-project/OpenSearch/blob/0eee9a9403e0338efde497da33fda87c39093b2c/server/src/main/java/org/opensearch/rest/BaseRestHandler.java#L103

src/main/java/org/opensearch/jobscheduler/rest/action/RestReleaseLockAction.java
```diff
@@ -0,0 +1,140 @@
...
+import org.opensearch.rest.BaseRestHandler;
+import org.opensearch.rest.BytesRestResponse;
+import org.opensearch.rest.RestRequest;
+import static org.opensearch.rest.RestRequest.Method.PUT;
+import org.opensearch.rest.RestStatus;
+
+public class RestReleaseLockAction extends BaseRestHandler {
+
+    public static final String RELEASE_LOCK_ACTION = "release_lock_action";
+    public static final String LOCK_ID = "lock_id";
```
saratvemulapalli, line 37: Lets import this from `org.opensearch.jobscheduler.spi.LockModel.LOCK_ID`.

src/main/java/org/opensearch/jobscheduler/rest/action/RestReleaseLockAction.java
```diff
@@ -0,0 +1,140 @@
...
+            findInProgressFuture.completeExceptionally(exception);
+        }));
+
+        try {
+            findInProgressFuture.orTimeout(JobDetailsService.TIME_OUT_FOR_REQUEST, TimeUnit.SECONDS).join();
+        } catch (CompletionException e) {
+            if (e.getCause() instanceof TimeoutException) {
+                logger.info(" Request timed out with an exception ", e);
+            } else {
+                throw e;
```
saratvemulapalli, line 79: Hmm.. do we need the else ? And throw the exception? 
The catch all at 87 will anyway take care of it.

src/main/java/org/opensearch/jobscheduler/rest/action/RestReleaseLockAction.java
```diff
@@ -0,0 +1,139 @@
...
+                    logger.error("Releasing lock failed with an exception", e);
+                    releaseLockInProgressFuture.completeExceptionally(e);
+                }
+            });
+
+            try {
+                releaseLockInProgressFuture.orTimeout(JobDetailsService.TIME_OUT_FOR_REQUEST, TimeUnit.SECONDS);
+            } catch (CompletionException e) {
+                if (e.getCause() instanceof TimeoutException) {
+                    logger.info(" Request timed out with an exception ", e);
```
joshpalis, line 104: lets use logger.error to log exceptions

src/main/java/org/opensearch/jobscheduler/rest/action/RestReleaseLockAction.java
```diff
@@ -0,0 +1,139 @@
...
+                } else {
+                    throw e;
+                }
+            } catch (Exception e) {
+                logger.info(" Could not release lock with " + releaseLock.getLockId() + " due to exception ", e);
```
joshpalis, line 109: here as well

src/main/java/org/opensearch/jobscheduler/rest/action/RestReleaseLockAction.java
```diff
@@ -0,0 +1,139 @@
...
+            logger.error("Could not find lock model with lockId " + lockId, exception);
+            findInProgressFuture.completeExceptionally(exception);
+        }));
+
+        LockModel releaseLock = null;
+        try {
+            releaseLock = findInProgressFuture.orTimeout(JobDetailsService.TIME_OUT_FOR_REQUEST, TimeUnit.SECONDS).get();
+        } catch (CompletionException e) {
+            if (e.getCause() instanceof TimeoutException) {
+                logger.info(" Request timed out with an exception ", e);
```
joshpalis, line 77: and here

src/main/java/org/opensearch/jobscheduler/rest/action/RestReleaseLockAction.java
```diff
@@ -0,0 +1,139 @@
...
+            } else {
+                throw e;
+            }
+        } catch (Exception e) {
+            logger.info(" Could not find lock model with lockId due to exception ", e);
```
joshpalis, line 82: also here

src/main/java/org/opensearch/jobscheduler/rest/action/RestReleaseLockAction.java
```diff
@@ -0,0 +1,140 @@
...
+        try {
+            findInProgressFuture.orTimeout(JobDetailsService.TIME_OUT_FOR_REQUEST, TimeUnit.SECONDS).join();
+        } catch (CompletionException e) {
+            if (e.getCause() instanceof TimeoutException) {
+                logger.info(" Request timed out with an exception ", e);
+            } else {
+                throw e;
+            }
+        } catch (Exception e) {
+            logger.info(" Could not find lock model with lockId due to exception ", e);
```
saratvemulapalli: Have this log at debug level?[1] Info seems like a lot.
 

[1] https://www.tutorialspoint.com/log4j/log4j_logging_levels.htm

joshpalis (commented):
Second pass. The modifications made to the Rest actions to use completableFuture.get() looks good to me, thanks @dbwiddis for the suggestion. Just had a few minor nits

dbwiddis (dismissed):

src/main/java/org/opensearch/jobscheduler/rest/action/RestReleaseLockAction.java
vibrantvarun: ACK

src/main/java/org/opensearch/jobscheduler/rest/action/RestReleaseLockAction.java
vibrantvarun: ACk

src/main/java/org/opensearch/jobscheduler/rest/action/RestReleaseLockAction.java
vibrantvarun: ACK

src/main/java/org/opensearch/jobscheduler/rest/action/RestReleaseLockAction.java
vibrantvarun: ACK

src/main/java/org/opensearch/jobscheduler/rest/action/RestReleaseLockAction.java
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        return channel -> {
+            XContentBuilder builder = channel.newBuilder();
+            RestStatus restStatus = RestStatus.OK;
+            String restResponseString = releaseResponse ? "success" : "failed";
+            BytesRestResponse bytesRestResponse;
+            try {
+                builder.startObject();
+                builder.field("response", restResponseString);
```
saratvemulapalli, line 126: I would re-name the field to `release-lock` to be more explicit when the client reads it. `response` is generic and it bring ambiguity. 
```suggestion
                builder.field("release-lock", restResponseString);
```

joshpalis (approved):
Thanks for the changes @vibrantvarun

saratvemulapalli (approved):
Overall the changes look good to me. 
Few minor comments..

src/main/java/org/opensearch/jobscheduler/rest/action/RestGetLockAction.java
vibrantvarun: yes, because if it is not public then I cannot call it from Test class

src/main/java/org/opensearch/jobscheduler/rest/action/RestReleaseLockAction.java
vibrantvarun: ACK

src/main/java/org/opensearch/jobscheduler/rest/action/RestReleaseLockAction.java
vibrantvarun: @dbwiddis suggested to catch time out exception seperately

src/main/java/org/opensearch/jobscheduler/rest/action/RestReleaseLockAction.java
vibrantvarun: ACk

src/main/java/org/opensearch/jobscheduler/rest/action/RestReleaseLockAction.java
vibrantvarun: ACK

src/main/java/org/opensearch/jobscheduler/rest/action/RestReleaseLockAction.java
saratvemulapalli: Ok if so we need need handle this cleanly, would it drop the rest request ? 
Is there a test which tests this case?

vibrantvarun: No as of now there is no test for this

src/main/java/org/opensearch/jobscheduler/rest/action/RestGetLockAction.java
dbwiddis: you could make it package private and available by tests

src/main/java/org/opensearch/jobscheduler/rest/action/RestReleaseLockAction.java
saratvemulapalli: Lets add one.

dbwiddis: OK, I think I may have been confused before and reviewing a PR on opensearch cleared it up.  We do need to clean up this exception handling.  Take a look at `ExternsionsManager` and try to follow this pattern:
```java
} catch (CompletionException e) {
    if (e.getCause() instanceof TimeoutException) {
        logger.info("No response from extension to request.");
    }
    if (e.getCause() instanceof RuntimeException) {
        throw (RuntimeException) e.getCause();
    } else if (e.getCause() instanceof Error) {
        throw (Error) e.getCause();
    } else {
        throw new RuntimeException(e.getCause());
    }
}
```

src/main/java/org/opensearch/jobscheduler/rest/action/VisibleForTesting.java
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.jobscheduler.rest.action;
+
+public @interface VisibleForTesting {
```
dbwiddis, line 11: Doesn't this already exist?
 - `import com.google.common.annotations.VisibleForTesting;` if you have that dependency
 - `import org.opensearch.jobscheduler.utils.VisibleForTesting;` otherwise

dbwiddis (approved):
LGTM but you're duplicating that annotation from elsewhere

src/main/java/org/opensearch/jobscheduler/rest/action/VisibleForTesting.java
vibrantvarun: it does have but it only works when when it is in the same package

src/main/java/org/opensearch/jobscheduler/rest/action/RestGetLockAction.java
```diff
@@ -82,36 +79,39 @@ protected RestChannelConsumer prepareRequest(RestRequest restRequest, NodeClient
...
-                logger.info("Could not acquire lock with ID : " + jobId, e);
-                inProgressFuture.completeExceptionally(e);
-            }
-        });
+        lockService.acquireLockWithId(
+            jobIndexName,
+            lockDurationSeconds,
+            jobId,
+            ActionListener.wrap(lockModel -> { inProgressFuture.complete(lockModel); }, exception -> {
+                logger.info("Could not acquire lock with ID : " + jobId, exception);
```
joshpalis, line 87: Lets make this logger.error

joshpalis (changes requested):
Doing another pass, some minor nits and a request to add guardrails to the releaseLock API

src/main/java/org/opensearch/jobscheduler/rest/action/RestReleaseLockAction.java
```diff
@@ -0,0 +1,144 @@
...
+
+    @Override
+    public RestChannelConsumer prepareRequest(RestRequest restRequest, NodeClient nodeClient) throws IOException {
+        String lockId = restRequest.param(LockModel.LOCK_ID);
+        if (lockId == null || lockId.isEmpty()) {
+            throw new IOException("lockId cannot be null or empty");
+        }
+
+        CompletableFuture<LockModel> findInProgressFuture = new CompletableFuture<>();
+        lockService.findLock(lockId, ActionListener.wrap(lock -> { findInProgressFuture.complete(lock); }, exception -> {
```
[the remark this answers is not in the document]
vibrantvarun: ACk

src/main/java/org/opensearch/jobscheduler/rest/action/RestGetLockAction.java
vibrantvarun: ACK

src/main/java/org/opensearch/jobscheduler/rest/action/RestGetLockAction.java
```diff
@@ -82,36 +79,39 @@ protected RestChannelConsumer prepareRequest(RestRequest restRequest, NodeClient
...
-                throw e;
+                throw new RuntimeException(e.getCause());
             }
-        } catch (Exception e) {
-            logger.error(" Could not process acquire lock request due to exception ", e);
         }
 
         return channel -> {
             BytesRestResponse bytesRestResponse;
+            LockModel lockModelResponseHolder = null;
```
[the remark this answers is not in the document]
vibrantvarun: ACK

joshpalis (approved):
Thanks for the changes @vibrantvarun

saratvemulapalli (approved):

Issue #305, opened 2023-01-20
[FEATURE] Make the job scheduler lock index a system index


bowenlan-amzn:
The job scheduler lock index (`.opendistro-job-scheduler-lock`) is used fully internally and shouldn't be changed by user.
If user inadvertently change this index, it could cause jobs to not running anymore.
Making it a system index can protect it from being changed by user.

cwperks:
I don't see any instances of `stashContext` in the code so that would need to be added before the system index can be added in the security plugin. Can I take on this task?

PR #347, opened 2023-03-22
Wrap interactions with .opendistro-job-scheduler-lock in ThreadContext.stashContext to ensure JS can read and write to the index

cwperks:
### Description

This PR wraps every instance of JS interacting with `.opendistro-job-scheduler-lock` inside of `stashContext` which allows the plugin to interact with system indices that have special protections. System indices cannot be inadvertently created, deleted or modified by even the admin user of the cluster which protects the cluster from getting into a corrupted state. 

~~This also makes JS extend SystemIndexPlugin which registers it with core a system index.~~ Done separately in https://github.com/opensearch-project/job-scheduler/pull/474
 
### Issues Resolved

https://github.com/opensearch-project/job-scheduler/issues/305
 
### Check List
  - [ ] All tests pass
- [ ] Commits are signed per the DCO using --signoff 


cwperks:
The failing check `./gradlew :opensearch-job-scheduler-sample-extension:integTest` ran successfully on my local machine.

saratvemulapalli:
@cwperks how does this work if the index exists in the system in older version of the distribution and now moving to system index in version? How does security/OpenSearch handle it? 
I dont see problems with JS as the index was a normal index and now the newer version of plugin has a way to access the system index.

cwperks:
@saratvemulapalli I will create the index as a normal index before installing JS and then install JS with this change and verify that it works as expected. I believe it can be converted without issue, but I'm not positive. Do you know where I can find information on how core treats system indices? I see logic that auto creates system indices if they do not exists, but is there anything else that is special about them in core? In the security plugin what is special about system indices is that no one can delete them unless you connect using admin cert and it matches the admin_dn. 

I have not been able to replicate the build issues on my dev machine. I wanted to go through the plugin developer experience of working with security so this looked like a good issue to try.

cwperks:
@saratvemulapalli Its' possible to give the index system index protection via the security plugin, and ensure that the plugin can write to it via a PR like this, but I can't currently find a way to make it a system index without steps to migrate like deleting the index and having it be recreated. Making it a formal system index involves extending `SystemIndexPlugin` and overwriting `getSystemIndexDescriptors` and that comes with some features like having the index automatically be created. 

I will close this PR if its not a change that's desired. I wanted to take an issue to get more familiar with JS without taking one of the issues allocated for CCI.

cwperks:
Re-opening and rebasing this PR. The lock index was marked as a system index using the extension point SystemIndexPlugin.getSystemIndexDescriptor in https://github.com/opensearch-project/job-scheduler/pull/474 and maintained backward compatibility. 

This PR surrounds calls to the lock index in a `ThreadContext.stashContext` block to ensure that job scheduler can interact with the index in a trusted context. 

Resolves errors seen in https://github.com/opensearch-project/job-scheduler/issues/492 after the lock index was added in the security plugin's system index list.

cwperks:
@prudhvigodithi Can you review this change?

prudhvigodithi:
Hey @cwperks thanks for the contribution, Coming from this [comment](https://github.com/opensearch-project/job-scheduler/issues/492#issuecomment-1716319011) I understand its necessary to stash the thread context in JobScheduler for every CRUD operation on the lock index which is done part of this PR, after this PR I assume (please correct me if I'm wrong) the next step is also to update the the [SYSTEM_INDICES](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java#L49C31-L49C45) list to add `.opendistro-job-scheduler-lock` which was reverted as part of this [PR](https://github.com/opensearch-project/security/pull/3237/files). Once this is merged is it required to modify the integration tests (with security) for other plugins that use the job-scheduler?
@getsaurabh02

cwperks:
@prudhvigodithi That's correct. As part of https://github.com/opensearch-project/security/issues/4439, I plan to make a change to feed the registry of system indices from core -> security plugin to avoid having to track them in a separate list in the security plugin. The job scheduler's lock index is one of the system indices that's already included in the registry and needs this change to ensure job-scheduler can interact with the index as desired.

This is actually a bug in job-scheduler atm where it will either use the current authenticated user or the roles that were injected into the threadcontext when determining whether a write operation can happen to the lock index. This PR will ensure that job-scheduler will always be able to interact with the index.

prudhvigodithi:
Thanks @cwperks LGTM, since you are working on adding some unit tests, you can either go in a new PR and add them here.

prudhvigodithi (approved):

cwperks:
> Just curious are we referring to the NamedXContentRegistry and index .opendistro-job-scheduler-lock?

The registry of system indices meaning that this plugin implements SystemIndexPlugin.getSystemIndexDescriptors(). 

I'm working on a change to provide protection by default to all plugins that implement SystemIndexPlugin.getSystemIndexDescriptors(), without having to submit a PR to the security repo to add it in the list tracked here: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java#L49-L80

PR #349, opened 2023-03-24
[Extensions] Fixes serde logic for proxy scheduled jobrunner/parser requests/responses

joshpalis:
### Description
Coming from modifications to the `ExtensionProxyAction` workflow [here](https://github.com/opensearch-project/OpenSearch/pull/6734).

The `JobDetailsService` used to facilitate job execution for extensions creates proxy `ScheduledJobRunner` and `ScheduledJobParser` objects that utilize the node client to execute remote extension actions. 

Here is an example of the new workflow : 

1. An extension registers its Job Details and Job Scheduler creates a `ScheduledJobProvider`, which contains a `ScheduledJobParser` that delegates parsing of the extension job index entry to the extension by executing the extension parse action
2. Job Scheduler is notified of an index operation on the extension job index, pulls that entry and it's proxy `ScheduledJobParser` invokes [ExtensionProxyAction](https://github.com/opensearch-project/OpenSearch/blob/9febe1041724c84409b18d2ca4f6946f5616e99c/server/src/main/java/org/opensearch/extensions/action/ExtensionProxyAction.java#L18) to send this to the extension to parse
3. The `ExtensionProxyJobAction` attaches the fully qualified class name of the `ExtensionProxyAction` as bytes in the `responseByte` field, along with the bytes provided by the inner `JobParserRequest` that contains the job index entry source
4. This request is sent to the SDK's [REQUEST_EXTENSION_HANDLE_TRANSPORT_ACTION handler](https://github.com/opensearch-project/opensearch-sdk-java/blob/ebc684ae223e87f1e05712e086fc1dab911a3b5f/src/main/java/org/opensearch/sdk/ExtensionsRunner.java#L407)
5. This request is then handled by the `ExtensionActionRequestHandler` [here](https://github.com/opensearch-project/opensearch-sdk-java/blob/ebc684ae223e87f1e05712e086fc1dab911a3b5f/src/main/java/org/opensearch/sdk/handlers/ExtensionActionRequestHandler.java#L59), which delegates the request handling to the `handleRemoteExtensionActionRequest` [method](https://github.com/opensearch-project/opensearch-sdk-java/blob/ebc684ae223e87f1e05712e086fc1dab911a3b5f/src/main/java/org/opensearch/sdk/handlers/ExtensionActionRequestHandler.java#LL73C42-L73C76).
6. Within this method, the `requestBytes` are processed and the request class name is identified [here](https://github.com/opensearch-project/opensearch-sdk-java/blob/ebc684ae223e87f1e05712e086fc1dab911a3b5f/src/main/java/org/opensearch/sdk/handlers/ExtensionActionRequestHandler.java#L87) which is used to identify and create the constructor for this particular `ActionRequest`
7. The action is then executed using the `SDKClient` and this request is forwarded to the corresponding `TransportAction`
8. The response is then sent back to Job Scheduler

This PR ensures that request bytes for the `ExtensionActionRequest` sent from Job Scheduler are in the correct format to be processed by the SDK.

 
### Check List
  - [X] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/jobscheduler/transport/request/ExtensionJobActionRequest.java
+++ b/src/main/java/org/opensearch/jobscheduler/transport/request/ExtensionJobActionRequest.java
@@ -9,15 +9,22 @@
 package org.opensearch.jobscheduler.transport.request;
 
 import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+
+import org.opensearch.common.bytes.BytesReference;
+import org.opensearch.common.io.stream.BytesStreamOutput;
 import org.opensearch.common.io.stream.Writeable;
 import org.opensearch.extensions.action.ExtensionActionRequest;
-import org.opensearch.jobscheduler.utils.JobDetailsService;
 
 /**
  * Request to extensions to invoke a job action, converts request params to a byte array
  *
  */
 public class ExtensionJobActionRequest<T extends Writeable> extends ExtensionActionRequest {
+
+    public static final byte UNIT_SEPARATOR = (byte) '\u001F';
 
     /**
      * Instantiates a new ExtensionJobActionRequest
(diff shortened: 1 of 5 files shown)


src/main/java/org/opensearch/jobscheduler/transport/request/ExtensionJobActionRequest.java
```diff
@@ -27,7 +34,63 @@
...
+    }
+
+    /**
+     * Finds the index of the specified byte value within the given byte array
+     *
+     * @param bytes the byte array to process
+     * @param value the byte to identify index of
+     * @return the index of the byte value
+     */
+    private static int indexOf(byte[] bytes, byte value) {
```
dbwiddis, line 47: I wonder if we should make this a util method in ExtensionTransportAction

src/main/java/org/opensearch/jobscheduler/transport/request/ExtensionJobActionRequest.java
```diff
@@ -27,7 +34,63 @@
...
+    }
+
+    /**
+     * Trims off the fully qualified request class name bytes and null byte from the ExtensionActionRequest requestBytes
+     *
+     * @param requestBytes the request bytes of an ExtensionActionRequest
+     * @return the trimmed array of bytes
+     */
+    public static byte[] trimRequestBytes(byte[] requestBytes) {
+        int nullPos = indexOf(requestBytes, ExtensionJobActionRequest.UNIT_SEPARATOR);
```
dbwiddis, line 63: Note that `nullPos` was left over from when this was a null byte, and maybe should just be a `pos`.

src/main/java/org/opensearch/jobscheduler/transport/request/ExtensionJobActionRequest.java
```diff
@@ -27,7 +34,63 @@
...
+     * @param actionParams the action parameters to be serialized
+     * @throws IOException if serialization fails
+     * @return the byte array of the parameters
+     */
+    private static <T extends Writeable> byte[] convertParamsToBytes(T actionParams) throws IOException {
+
+        // Write inner request to output stream and convert to byte array
+        BytesStreamOutput out = new BytesStreamOutput();
+        actionParams.writeTo(out);
+        out.flush();
```
dbwiddis, line 80: I used to always do this, but this is the method being invoked:
```java
    @Override
    public void flush() {
        // nothing to do
    }
```

src/main/java/org/opensearch/jobscheduler/transport/request/ExtensionJobActionRequest.java
```diff
@@ -27,7 +34,63 @@
...
+
+    /**
+     * Converts an object of type T that extends {@link Writeable} into a byte array and prepends the fully qualified class name bytes
+     *
+     * @param <T> a class that extends writeable
+     * @param actionParams the action parameters to be serialized
+     * @throws IOException if serialization fails
+     * @return the byte array of the parameters
+     */
+    private static <T extends Writeable> byte[] convertParamsToBytes(T actionParams) throws IOException {
```
dbwiddis, line 75: I wonder if this should be a util method on ExtensionActionRequest or ExtensionTransportAction

src/main/java/org/opensearch/jobscheduler/transport/response/ExtensionJobActionResponse.java
```diff
@@ -26,7 +28,25 @@
...
+
+    /**
+     * Takes in an object of type T that extends {@link Writeable} and converts the writeable fields to a byte array
+     *
+     * @param <T> a class that extends writeable
+     * @param actionParams the action parameters to be serialized
+     * @throws IOException if serialization fails
+     * @return the byte array of the parameters
+     */
+    private static <T extends Writeable> byte[] convertParamsToBytes(T actionParams) throws IOException {
```
dbwiddis, line 42: Deja vu.  We definitely should find a central location for these util classes that both Extensions and Plugins can use.

dbwiddis (approved):
LGTM for working code.

But I see multiple copies of the same utility code for parsing byte arrays associated with these remote calls, and they should probably find a home in ExtensionTransportAction or maybe a new ExtensionTransportUtil class.

OK to merge this as is but if anyone else thinks like I do let's make an issue to do this.

joshpalis:
@dbwiddis I agree, lets move these methods over to `ExtensionActionRequest`/`ExtensionTransportAction`. For now lets merge this change in until https://github.com/opensearch-project/opensearch-sdk-java/issues/607 is resolved. Once those utilities are available then I'll make the change to Job Scheduler to use those instead.

vibrantvarun:
LGTM!

vibrantvarun (approved):

PR #409, opened 2023-06-30
fix: time out of range

fudongyingluck:
We found that when current time greater than `nextExecutionTime`, the `TimeValue` in `threadPool.schedule` will throw an `IllegalArgumentException` as following
```[2023-06-06T22:10:04,144][ERROR][o.e.b.ElasticsearchUncaughtExceptionHandler] [es-data-y850og2ns2zzcu5n-5] uncaught exception in thread [elasticsearch[es-data-y850og2ns2zzcu5n-5][open_distro_job_scheduler][T#8]]
java.lang.IllegalArgumentException: duration cannot be negative, was given [-2965077933106]
        at org.elasticsearch.common.unit.TimeValue.<init>(TimeValue.java:52) ~[elasticsearch-core-7.10.2.jar:7.10.2]
        at com.amazon.opendistroforelasticsearch.jobscheduler.scheduler.JobScheduler.reschedule(JobScheduler.java:190) ~[?:?]
        at com.amazon.opendistroforelasticsearch.jobscheduler.scheduler.JobScheduler.lambda$reschedule$0(JobScheduler.java:177) ~[?:?]
        at org.elasticsearch.common.util.concurrent.ThreadContext$ContextPreservingRunnable.run(ThreadContext.java:684) ~[elasticsearch-7.10.2.jar:7.10.2]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130) ~[?:?]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630) ~[?:?]
        at java.lang.Thread.run(Thread.java:832) [?:?]
```

Then the job will not be scheduled anymore. 
This change fixes this, by setting the `nextExecutionTime` to current time.

Thanks my colleague @kkewwei solve this out.

Signed-off-by: fudongying <[fudongying@bytedance.com](mailto:fudongying@bytedance.com)>
Signed-off-by: kewei.11 <[kewei.11@bytedance.com](mailto:kewei.11@bytedance.com)>

### DIFF
--- a/src/main/java/org/opensearch/jobscheduler/scheduler/JobScheduler.java
+++ b/src/main/java/org/opensearch/jobscheduler/scheduler/JobScheduler.java
@@ -148,7 +148,18 @@ public class JobScheduler {
             log.info("No next execution time for job {}", jobParameter.getName());
             return true;
         }
-        Duration duration = Duration.between(this.clock.instant(), nextExecutionTime);
+        Instant now = this.clock.instant();
+        Duration duration = Duration.between(now, nextExecutionTime);
+        if (duration.toNanos() < 0) {
+            log.info(
+                "job {} expect time: {} < current time: {}, set next excute time to current",
+                jobParameter.getName(),
+                nextExecutionTime.toEpochMilli(),
+                now.toEpochMilli()
+            );
+            nextExecutionTime = now;
+            duration = Duration.ZERO;
+        }
 
         // Too many jobs start at the same time point will bring burst. Add random jitter delay to spread out load.
         // Example, if interval is 10 minutes, jitter is 0.6, next job run will be randomly delayed by 0 to 10*0.6 minutes.
--- a/src/main/java/org/opensearch/jobscheduler/scheduler/JobScheduler.java
+++ b/src/main/java/org/opensearch/jobscheduler/scheduler/JobScheduler.java
@@ -119,7 +119,7 @@ public class JobScheduler {
         jobInfo.setExpectedPreviousExecutionTime(null);
         Scheduler.ScheduledCancellable scheduledCancellable = jobInfo.getScheduledCancellable();
 
-        if (scheduledCancellable != null && scheduledCancellable.cancel() == false) {
+        if (scheduledCancellable != null && !scheduledCancellable.cancel()) {
             return false;
         }
         this.scheduledJobInfo.removeJob(indexName, id);
@@ -147,9 +147,9 @@ public class JobScheduler {
         }
         Instant now = this.clock.instant();
         Duration duration = Duration.between(now, nextExecutionTime);
-        if (duration.toNanos() < 0) {
+        if (duration.isNegative()) {
             log.info(
-                "job {} expect time: {} < current time: {}, set next excute time to current",
+                "job {} expected time: {} < current time: {}, setting next execute time to current",
                 jobParameter.getName(),
                 nextExecutionTime.toEpochMilli(),
                 now.toEpochMilli()
--- a/src/main/java/org/opensearch/jobscheduler/scheduler/JobScheduler.java
+++ b/src/main/java/org/opensearch/jobscheduler/scheduler/JobScheduler.java
@@ -119,13 +119,10 @@ public class JobScheduler {
         jobInfo.setExpectedPreviousExecutionTime(null);
         Scheduler.ScheduledCancellable scheduledCancellable = jobInfo.getScheduledCancellable();
 
-        if (scheduledCancellable != null) {
-            if (scheduledCancellable.cancel()) {
-                this.scheduledJobInfo.removeJob(indexName, id);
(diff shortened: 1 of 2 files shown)


joshpalis:
Thanks for raising this PR @fudongyingluck. Checks are failing due to stale artifacts, will wait to re-run checks until the next 2.9.0 build is successful
```
* What went wrong:
Could not determine the dependencies of task ':opensearch-job-scheduler-sample-extension:jobSchedulerBwcCluster#fullRestartClusterTask'.
> Server returned HTTP response code: 403 for URL: https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/2.9.0/8039/linux/x64/tar/builds/opensearch/plugins/opensearch-job-scheduler-2.9.0.0.zip
```

fudongyingluck:
@joshpalis Since we found that dirty data stale in the memory after exception. If the index migrates back to this node after we migrate it to another, then the job loses again. So we add the second commit to deal with this condition.

fudongyingluck:
> Thanks for raising this PR @fudongyingluck. Checks are failing due to stale artifacts, will wait to re-run checks until the next 2.9.0 build is successful
> 
> ```
> * What went wrong:
> Could not determine the dependencies of task ':opensearch-job-scheduler-sample-extension:jobSchedulerBwcCluster#fullRestartClusterTask'.
> > Server returned HTTP response code: 403 for URL: https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/2.9.0/8039/linux/x64/tar/builds/opensearch/plugins/opensearch-job-scheduler-2.9.0.0.zip
> ```

@joshpalis Checks seem successful now. Is it convenient for you to review those codes ? Really thanks for your time ~

src/main/java/org/opensearch/jobscheduler/scheduler/JobScheduler.java
```diff
@@ -119,13 +119,10 @@ public boolean deschedule(String indexName, String id) {
         jobInfo.setExpectedPreviousExecutionTime(null);
         Scheduler.ScheduledCancellable scheduledCancellable = jobInfo.getScheduledCancellable();
 
-        if (scheduledCancellable != null) {
-            if (scheduledCancellable.cancel()) {
-                this.scheduledJobInfo.removeJob(indexName, id);
-            } else {
-                return false;
-            }
+        if (scheduledCancellable != null && scheduledCancellable.cancel() == false) {
```
dbwiddis, line 122:
```suggestion
        if (scheduledCancellable != null && !scheduledCancellable.cancel()) {
```

src/main/java/org/opensearch/jobscheduler/scheduler/JobScheduler.java
```diff
@@ -148,7 +145,18 @@ boolean reschedule(
             log.info("No next execution time for job {}", jobParameter.getName());
             return true;
         }
-        Duration duration = Duration.between(this.clock.instant(), nextExecutionTime);
+        Instant now = this.clock.instant();
+        Duration duration = Duration.between(now, nextExecutionTime);
+        if (duration.toNanos() < 0) {
```
dbwiddis, line 150:
```suggestion
        if (duration.isNegative()) {
```

src/main/java/org/opensearch/jobscheduler/scheduler/JobScheduler.java
```diff
@@ -148,7 +145,18 @@ boolean reschedule(
...
+            log.info(
+                "job {} expect time: {} < current time: {}, set next excute time to current",
```
dbwiddis, line 152:
```suggestion
                "job {} expected time: {} < current time: {}, setting next execute time to current",
```

dbwiddis (commented):
Thanks for this fix!  

In the future, it would be helpful to create an issue reporting the details of the bug and then reference that issue in the PR. In case there's a need to discuss appropriate fixes for the bug that the PR might not handle it keeps the discussions separate.

Fix LGTM with a few nits.

fudongyingluck:
@dbwiddis  Really thanks for your time and advice. I'll create a bug issue next time. And The code is changed as your comments in the latest commit.

fudongyingluck:
@cwperks Good question. We also feel curious when the job disappears until we found logs. The thing is the cloud service which the ES k8s instance runs on, is unavailable for some time.  Then the ES instance seems not to run at those times, I don't know how this happened.  After about 30m, the ES instance reruns again, and the exception occurs.
I know we should fix the ES instance stalled problem, but this seems more complex, and the online ES instance can't restart for some stability reason.  To avoid the problem occurring next time, I raise this PR.

cwperks:
@fudongyingluck Thank you for the context!

cwperks (approved):

joshpalis (approved):

PR #474, opened 2023-08-24
Converts .opendistro-job-scheduler-lock index into a system index

joshpalis:
### Description
Actioning two issues to convert the `.opendistro-job-scheduler-lock` index into a system index.
This PR makes the following modifications :
- makes `JobSchedulerPlugin` implement the `SystemIndexPlugin` and overrides the `getSystemIndexDescriptors` extension point
- ~wraps each instance in which the plugin interacts with the system index within the `LockService` to stash the threadcontext prior to executing a CRUD operation. From a security perspective this allows the plugin to go into a local trusted mode in which they are allowed to interact with the system index~

Next steps :
- Modify the installation scripts ([sh](https://github.com/opensearch-project/security/blob/9a7e517abd9d4d0e301f918adc65cb82b8f14d88/tools/install_demo_configuration.sh#L386), [bat](https://github.com/opensearch-project/security/blob/9a7e517abd9d4d0e301f918adc65cb82b8f14d88/tools/install_demo_configuration.bat#L318C18-L318C18)) for unix/windows systems to add the system index name to the list of system indices that the security plugin maintains
- Companion PR : https://github.com/opensearch-project/security/pull/3237
 
### Issues Resolved
https://github.com/opensearch-project/job-scheduler/issues/466 - System Index Usage Verification
https://github.com/opensearch-project/job-scheduler/issues/305 
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/spi/src/main/java/org/opensearch/jobscheduler/spi/utils/LockService.java
+++ b/spi/src/main/java/org/opensearch/jobscheduler/spi/utils/LockService.java
@@ -8,7 +8,6 @@
  */
 package org.opensearch.jobscheduler.spi.utils;
 
-import org.opensearch.common.util.concurrent.ThreadContext;
 import org.opensearch.jobscheduler.spi.JobExecutionContext;
 import org.opensearch.jobscheduler.spi.LockModel;
 import org.opensearch.jobscheduler.spi.ScheduledJobParameter;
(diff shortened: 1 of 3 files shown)


spi/src/main/java/org/opensearch/jobscheduler/spi/utils/LockService.java
```diff
@@ -44,7 +45,7 @@
 
 public final class LockService {
     private static final Logger logger = LogManager.getLogger(LockService.class);
-    private static final String LOCK_INDEX_NAME = ".opendistro-job-scheduler-lock";
```
peternied, line 47: While we are making a change of this nature can we rename the index to opensearch?

peternied: Resolved this on the security plugin change to add the system index - since the the index name isn't change at all, it would introduce much more risk for little gain.

joshpalis:
Interestingly the SampleJobRunnerIT tests are failing consistently, but pass locally. I will continue to look into this

spi/src/main/java/org/opensearch/jobscheduler/spi/utils/LockService.java
```diff
@@ -220,7 +220,7 @@ private void updateLock(final LockModel updateLock, ActionListener<LockModel> li
             listener.onResponse(null);
         } catch (Exception e) {
             logger.error(e);
-            listener.onResponse(null);
+            listener.onFailure(e);
```
peternied, line 223: Should the line at 220 also be handled the same way?

joshpalis: This was my initial thought as well, but looking at the usage of `updateLock()` within `renewLock` and `releaseLock` it seems that its meant to return null on response for some cases. 

If stashing the thread context fails for any reason then we'll want to fail that directly.

vibrantvarun:
Tests are failing

cwperks:
It may be the case that JobScheduler does not need to wrap calls to interact with the lock system index inside of a `threadContext.stashContext { ... }` block, but I don't think it would hurt either.

Typically stashContext is used within plugins when handling _authenticated_ requests. (i.e. API Endpoint Handling) which JS does not do to my knowledge.

A good example is Anomaly Detection create detector.

1. User calls on endpoint to create detector and passes credentials with the request
2. The security plugin wraps the REST handler, verifies the credentials and authenticates the request. Authentication here means that the security plugin injects the User into the threadcontext
3. When AD wants to write a document to the detector system index it will stash the threadcontext (effectively switching from the context of the authenticated user into a new trusted context where it can do anything) and index the document into its system index containing the detector definitions.

If JS is always outside of an authenticated user context than stashing the threadcontext could be considered redundant, however, there are other avenues for plugins to inject roles/user into the threadcontext so stashing the context ensures that the block executing code has a clean slate.

joshpalis:
> It may be the case that JobScheduler does not need to wrap calls to interact with the lock system index inside of a threadContext.stashContext { ... } block, but I don't think it would hurt either.

I've observed that the integration tests would fail consistently after this change, at first I considered it might be a transitive error but after a few test runs it hasn't been resolved. I'll opt to revert the changes to stash the thread context

vibrantvarun (approved):

cwperks (approved):

PR #714, opened 2025-01-15
Make LockService an interface and replace usages of ThreadContext.stashContext

cwperks:
### Description

This PR removes direct usages of ThreadContext.stashContext and implements a new extension point in core (IdentityAwarePlugin) which assigns this plugin a subject to use for the purposes of executing transport actions outside of the authenticated user context.

With this subject, job-scheduler is permitted to access its own system indices without having to use `ThreadContext.stashContext` explicitly.

This PR also updates the sample plugin to enable testing with Security.

You can test with security in the sample plugin using a command like this:

```
./gradlew :opensearch-job-scheduler-sample-extension:integTest -Dhttps=true -Dsecurity.enabled=true  -Dtests.opensearch.secure=true -Dtests.opensearch.username=admin -Dtests.opensearch.password=admin -Duser=admin -Dpassword=admin --tests 'SampleJobRunnerRestIT' -i
```

### Related Issues

Resolves https://github.com/opensearch-project/opensearch-plugins/issues/238

### Check List


spi/src/main/java/org/opensearch/jobscheduler/spi/utils/LockService.java
```diff
@@ -220,18 +215,16 @@ private void updateLock(final LockModel updateLock, ActionListener<LockModel> li
...
             listener.onFailure(e);
         }
     }
 
     private void createLock(final LockModel tempLock, ActionListener<LockModel> listener) {
-        try (ThreadContext.StoredContext ignore = client.threadPool().getThreadContext().stashContext()) {
+        try {
             final IndexRequest request = new IndexRequest(LOCK_INDEX_NAME).id(tempLock.getLockId())
                 .source(tempLock.toXContent(XContentFactory.jsonBuilder(), ToXContent.EMPTY_PARAMS))
+                .setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE)
```
cwperks, line 227: Not sure why this is necessary here just yet.

cwperks: This is needed to resolve a race condition discovered in [GetLockMultiNodeRestIT.testGetLockRestAPI](https://github.com/opensearch-project/job-scheduler/blob/main/src/test/java/org/opensearch/jobscheduler/multinode/GetLockMultiNodeRestIT.java#L50-L81).  This test will create and release a lock 10 times in a loop. Without adding this RefreshPolicy, when the security plugin is installed it can fail to find the lock the was just created. By setting this RefreshPolicy to IMMEDIATE, we can guarantee that this newly created document is available for search (including GET actions)

src/main/java/org/opensearch/jobscheduler/JobSchedulerPlugin.java
```diff
@@ -100,8 +104,9 @@ public Map<String, ScheduledJobProvider> getIndexToJobProviders() {
 
     @Override
     public Collection<SystemIndexDescriptor> getSystemIndexDescriptors(Settings settings) {
-        return Collections.singletonList(
-            new SystemIndexDescriptor(LockService.LOCK_INDEX_NAME, "Stores lock documents used for plugin job execution")
+        return List.of(
+            new SystemIndexDescriptor(LockServiceImpl.LOCK_INDEX_NAME, "Stores lock documents used for plugin job execution"),
```
kaituo, line 108: Does referencing LockServiceImpl.LOCK_INDEX_NAME couples the plugin to the impl and leaks impl constants across the boundary? Should we keep fields/params typed as the interface and move the lock index name into SPI (e.g., LockService.LOCK_INDEX_NAME) or a small LockConstants in SPI.

cwperks: Yes, it does couple it to System Index implementation which is one of the reasons I'd like to change this to an interface so a caller knows the interface, but not necessarily how its implemented.

I had made changes across a few plugins to remove direct instantiation of the LockService, but didn't realize some plugins reference [LOCK_INDEX_NAME](https://github.com/opensearch-project/security-analytics/blob/96a5ca2670e7434e3fcd11c2f85c095d40f12e8d/src/main/java/org/opensearch/securityanalytics/threatIntel/service/SATIFSourceConfigService.java#L422-L459). I don't see a compelling reason why other plugins need to be aware of the name of this index and the SAP use-case looks repetitive as JS will release a lock for disabled jobs. Are you aware of any bug in SAP for why that code was introduced?

kaituo: I am not aware of any bug in SAP. Would your current change break their code?

kaituo: I meant to ask should we change back to LockService.LOCK_INDEX_NAME? In this way, you won't break their code and avoid coupling.

"Does referencing LockServiceImpl.LOCK_INDEX_NAME couples the plugin to the impl and leaks impl constants across the boundary? Should we keep fields/params typed as the interface and move the lock index name into SPI (e.g., LockService.LOCK_INDEX_NAME) or a small LockConstants in SPI."

cwperks: Removing direct reference to LOCK_INDEX_NAME in SAP reduces coupling. We can put the LOCK_INDEX_NAME back in the SPI as a public constant, but I think it would make more sense to hide that from plugins as an implementation detail. Plugins should get the interface they need to interact with from the SPI.

cwperks: FYI Created this PR in SAP repo: https://github.com/opensearch-project/security-analytics/pull/1577

kaituo (approved):

PR #773, opened 2025-05-07
remove guava dependency

jngz-es:
### Description
As a common extensible plugin, remove guava dependency to reduce jar hell or conflicts issues caused by guava in extending plugins like SQL, ML etc.

### Related Issues
https://github.com/opensearch-project/OpenSearch/issues/18113

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -183,22 +183,10 @@ repositories {
     maven { url "https://aws.oss.sonatype.org/content/repositories/snapshots" }
 }
 
-configurations {
-    all {
-        resolutionStrategy {
-            force 'com.google.guava:guava:32.1.3-jre'
-        }
-    }
-}
-
 dependencies {
     implementation project(path: ":${rootProject.name}-spi", configuration: 'shadow')
     testImplementation group: 'org.mockito', name: 'mockito-core', version: "${versions.mockito}"
     javaRestTestImplementation project.sourceSets.main.runtimeClasspath
-    //spotless
-    implementation('com.google.googlejavaformat:google-java-format:1.26.0') {
-        exclude group: 'com.google.guava'
-    }
 
     opensearchPlugin "org.opensearch.plugin:opensearch-security:${security_plugin_version}@zip"
 }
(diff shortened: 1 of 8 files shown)


build.gradle
```diff
@@ -193,8 +193,6 @@ configurations {
 
 dependencies {
     implementation project(path: ":${rootProject.name}-spi", configuration: 'shadow')
-    implementation group: 'com.google.guava', name: 'guava', version:'32.1.3-jre'
```
andrross, line 196: I think the whole section above this can go too:

```
configurations {
    all {
        resolutionStrategy {
            force 'com.google.guava:guava:32.1.3-jre'
        }
    }
}
```

Also what about the section below?

```
    //spotless
    implementation('com.google.googlejavaformat:google-java-format:1.26.0') {
        exclude group: 'com.google.guava'
    }
```

andrross:
This makes sense to me! The JDK now natively supports these immutable collections so no need to bring in guava just to get ImmutableList and ImmutableMap.

prudhvigodithi:
Thanks @jngz-es thanks for this change. Same as what Andrew mentioned, I have tested with `List.of` and behaves the same
```
List<String> immutable = List.of("a", "b", "c");
immutable.add("d"); // this throws exception.
```

build.gradle
jngz-es: Good catch, I am removing them as well.

kaituo (approved):

kaituo:
rerun CI

cwperks:
I support removing guava from this plugin, but also wanted to leave insight I found out about extensible plugins recently.

When pluginA extends pluginB, all of pluginBs deps are available to pluginA at runtime (**not** compile time)

Because of this, plugins like AD are able to specify [compileOnly dependencies on guava](https://github.com/opensearch-project/anomaly-detection/blob/main/build.gradle#L140-L141) and know that they are available at runtime because of their extension on JS plugin.

The key is to ensure there is a version catalog to keep these common deps in sync.

cwperks (approved):

cwperks:
@kaituo make sure to update AD from `compileOnly` to `implementation` for the guava deps now that this change is merged.

cwperks:
> @kaituo make sure to update AD from `compileOnly` to `implementation` for the guava deps now that this change is merged.

@kaituo I raised https://github.com/opensearch-project/anomaly-detection/pull/1473 to address this

PR #801, opened 2025-07-23
Move info about delay to the the schedule portion in List Jobs API

Jeremydupras:
### Description
Lock information will be gathered through a separate API and correlated in a future job scheduler dashboard. This PR removes the locks qualities from the getScheduledJobInfo transport action and from tests.

The linked hash set was added to ensure the JSON always comes out in the same order.
Delay was added into the schedule data structure.

### Related Issues
no issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/jobscheduler/transport/action/TransportGetScheduledInfoAction.java
+++ b/src/main/java/org/opensearch/jobscheduler/transport/action/TransportGetScheduledInfoAction.java
@@ -214,6 +214,13 @@ public class TransportGetScheduledInfoAction extends TransportNodesAction<
                                     jobDetails.put("schedule", scheduleMap);
                                 }
 
+                                jobDetails.put(
+                                    "lock_duration",
+                                    jobInfo.getJobParameter().getLockDurationSeconds() != null
+                                        ? jobInfo.getJobParameter().getLockDurationSeconds()
+                                        : "no_lock"
+                                );
+
                                 // Add jitter and lock duration
                                 jobDetails.put(
                                     "jitter",


src/main/java/org/opensearch/jobscheduler/transport/action/TransportGetScheduledInfoAction.java
```diff
@@ -192,37 +191,34 @@ protected GetScheduledInfoNodeResponse nodeOperation(GetScheduledInfoNodeRequest
...
-                                            : "none"
-                                    );
                                 }
 
                                 // Add jitter and lock duration
                                 jobDetails.put(
                                     "jitter",
                                     jobInfo.getJobParameter().getJitter() != null ? jobInfo.getJobParameter().getJitter() : "none"
                                 );
-                                jobDetails.put(
```
cwperks, line 219: Let's keep `lock_duration` in. At this level, `lock_duration` represents the maximum amount of time a lock will be held for after its acquired. If a job runs and fails to release a lock in time (for example if the job fails while running), then JS has a mechanism to release the lock so that its not help in perpetuity.

src/main/java/org/opensearch/jobscheduler/transport/action/TransportGetScheduledInfoAction.java
```diff
@@ -181,7 +180,7 @@ protected GetScheduledInfoNodeResponse nodeOperation(GetScheduledInfoNodeRequest
                                 if (jobInfo.getJobParameter().getSchedule() == null) {
                                     log.debug("Schedule for job {} does not exist.", jobId);
                                 } else {
-                                    Map<String, Object> scheduleMap = new HashMap<>();
+                                    Map<String, Object> scheduleMap = new LinkedHashMap<>();
```
cwperks, line 183: Any specific reason for this change? Is order of insertion important?

Jeremydupras: The this ensures that the JSON output is always in the same order. This makes reading it easier and allows for a more predictable output.

PR #806, opened 2025-08-06
adding descheduled jobs the the Jobs API call

Jeremydupras:
### Description
Description
The Jobs REST API currently only lists the scheduled jobs which have the fields "descheduled = false and enabled = true". This will allow the user to see all jobs in job scheduler whether or not they are scheduled in both cluster wide call and by_node call.

There is no difference in the API call. The differences can be seen in the Descheduled and Enabled Fields. Additionally, "last_execution_time", "last_expected_execution_time", and "next_expected_execution_time" are all set to null values when the job is set to enables = false. See the example JSON output below.

Lastly, when a job is deleted it is removed the scheduledjobs data structure and the corresponding lock is removed regardless of the jobs enabled status.

```
{
  "jobs": [
    {
      "job_type": "scheduler_sample_extension",
      "job_id": "jobid2",
      "index_name": ".scheduler_sample_extension",
      "name": "sample-job-it",
      "descheduled": true,
      "enabled": false,
      "enabled_time": "1970-07-23T01:37:43.063Z",
      "last_update_time": "1970-07-23T01:37:43.063Z",
      "last_execution_time": "none",
      "last_expected_execution_time": "none",
      "next_expected_execution_time": "none",
      "schedule": {
        "type": "interval",
        "start_time": "1970-07-23T01:37:43.063Z",
        "interval": 1,
        "unit": "Minutes",
        "delay": "none"
      },
      "lock_duration": 10,
      "jitter": "none"
    }
  ],
  "failures": [],
  "total_jobs": 1
}
```

### Check List

### DIFF
--- a/sample-extension-plugin/src/test/java/org/opensearch/jobscheduler/sampleextension/GetScheduledJobInfoIT.java
+++ b/sample-extension-plugin/src/test/java/org/opensearch/jobscheduler/sampleextension/GetScheduledJobInfoIT.java
@@ -29,9 +29,16 @@ import java.util.HashSet;
 
 public class GetScheduledJobInfoIT extends SampleExtensionIntegTestCase {
 
+    Set<String> expectedJobIds = Set.of("test-job-1", "test-job-2", "test-job-3", "test-job-4");
+
+    private SampleJobParameter jobParam1;
+    private SampleJobParameter jobParam2;
+    private SampleJobParameter jobParam3;
+    private SampleJobParameter jobParam4;
+
     @Before
     public void setupJobs() throws IOException, InterruptedException {
-        SampleJobParameter jobParam1 = new SampleJobParameter(
+        jobParam1 = new SampleJobParameter(
             "test-job-1",
             "Test Job 1",
             "test-index-1",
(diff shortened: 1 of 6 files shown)


sample-extension-plugin/src/test/java/org/opensearch/jobscheduler/sampleextension/GetScheduledJobInfoIT.java
```diff
@@ -160,38 +131,167 @@ public void testGetScheduledJobInfoByNode() throws IOException {
...
+        for (Map<String, Object> job : jobs) {
+            validateJobFields(job, true, false, actualJobIds);
+        }
+        assertEquals("All expected job IDs should be present", expectedJobIds, actualJobIds);
+    }
+
+    private void validateJobFields(
+        Map<String, Object> job,
+        boolean expectedEnabled,
+        boolean expectedDescheduled,
```
cwperks, line 268: `expectedDescheduled` is the opposite of `exceptedEnabled` in every single call. Do we need 2 separate params? When would they both be true or both be false?

PR #814, opened 2025-08-13
Job History Service

### LINKED ISSUE #808: [FEATURE] Job execution History index
**Is your feature request related to a problem?**
This feature would help with trouble shooting by allowing users to see the past execution history all jobs within job scheduler.

A job history log will consist of:
Job index - which plugin is running the job
job Id - which specific job within the index
start time
end time
job completion status - Success/Fail

**What solution would you like?**
Create a Job History Service that logs the jobs as they run. This would be stored in a system index and can be correlated with the jobs API to view job history.

**Do you have any additional context?**
Future additions could include other parameters like which node is running the job.

Jeremydupras:
### Description
Creates a history index that records when a job acquires a lock to execute. Upon completion the record is updated with a completion time. The History service is built into the lock service and completely within jobScheduler. 

Logic path
->Extension plugin tries to acquire lock
if the lock is acquired then there is a new record added to the index. If not the nothing is added.
->status codes within the record show if the job finished execution
->upon completion the Extension plugin releases the lock and the record is updated to show the job completed

Example of the index contents
```
{
  "took" : 2,
  "timed_out" : false,
  "_shards" : {
    "total" : 1,
    "successful" : 1,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 2,
      "relation" : "eq"
    },
    "max_score" : 1.0,
    "hits" : [
      {
        "_index" : ".opendistro-job-scheduler-history",
        "_id" : ".scheduler_sample_extension-jobid1-1755118872",
        "_score" : 1.0,
        "_source" : {
          "job_index_name" : ".scheduler_sample_extension",
          "job_id" : "jobid1",
          "start_time" : 1755118872,
          "completion_status" : 0,
          "end_time" : 1755118872
        }
      },
      {
        "_index" : ".opendistro-job-scheduler-history",
        "_id" : ".scheduler_sample_extension-jobid1-1755118932",
        "_score" : 1.0,
        "_source" : {
          "job_index_name" : ".scheduler_sample_extension",
          "job_id" : "jobid1",
          "start_time" : 1755118932,
          "completion_status" : 0,
          "end_time" : 1755118932
        }
      }
    ]
  }
}
```

### Related Issues
Resolves #808 

### Check List

### DIFF
--- a/spi/src/main/java/org/opensearch/jobscheduler/spi/JobExecutionContext.java
+++ b/spi/src/main/java/org/opensearch/jobscheduler/spi/JobExecutionContext.java
@@ -22,7 +22,6 @@ public class JobExecutionContext implements Writeable {
     private final LockService lockService;
     private final String jobIndexName;
     private final String jobId;
-    private Integer jobStatus;
 
     public JobExecutionContext(
         Instant expectedExecutionTime,


src/test/java/org/opensearch/jobscheduler/utils/JobDetailsServiceIT.java
```diff
@@ -326,6 +327,7 @@ private void compareExtensionJobParameters(
     public void testJobRunnerExtensionJobActionRequest() throws IOException {
 
         LockService lockService = new LockService(client(), this.clusterService);
+        JobHistoryService historyService = new JobHistoryService(client(), clusterService);
```
DarshitChanpura, line 330: this declaration doesn't seem to be used. If so, remove it?

spi/src/main/java/org/opensearch/jobscheduler/spi/StatusHistoryModel.java
```diff
@@ -0,0 +1,194 @@
...
+
+    public StatusHistoryModel(StreamInput in) throws IOException {
+        this(
+            in.readString(),
+            in.readString(),
+            in.readInstant(),
+            in.readOptionalInstant(),
+            in.readInt(),
+            SequenceNumbers.UNASSIGNED_SEQ_NO,
+            SequenceNumbers.UNASSIGNED_PRIMARY_TERM
```
DarshitChanpura, line 71: this should be `in.readLong()` since you write long to output stream.

spi/src/main/java/org/opensearch/jobscheduler/spi/utils/JobHistoryService.java
```diff
@@ -0,0 +1,263 @@
...
+import java.time.Instant;
+import java.util.Map;
+
+public class JobHistoryService {
+    private static final Logger logger = LogManager.getLogger(JobHistoryService.class);
+    public static final String JOB_HISTORY_INDEX_NAME = ".opendistro-job-scheduler-history";
+
+    private final Client client;
+    private final ClusterService clusterService;
+    final static Map<String, Object> INDEX_SETTINGS = Map.of("index.number_of_shards", 1, "index.auto_expand_replicas", "0-1");
```
DarshitChanpura, line 48: replicas should 0-all. restricting it to 1 will only create 1 replica. `all` creates replicas based on number of nodes.

spi/src/main/java/org/opensearch/jobscheduler/spi/utils/JobHistoryService.java
```diff
@@ -0,0 +1,263 @@
...
+                final CreateIndexRequest request = new CreateIndexRequest(JOB_HISTORY_INDEX_NAME).mapping(
+                    historyMapping(),
+                    (MediaType) XContentType.JSON
+                ).settings(INDEX_SETTINGS);
+                client.admin()
+                    .indices()
+                    .create(request, ActionListener.wrap(response -> listener.onResponse(response.isAcknowledged()), exception -> {
+                        if (exception instanceof ResourceAlreadyExistsException
+                            || exception.getCause() instanceof ResourceAlreadyExistsException) {
+                            listener.onResponse(true);
```
DarshitChanpura, line 87: Do we return true here to acknowledge that the index already exists?

Should we create job history index once, inside onNodeStarted method is plugin class? This will save us from calling createJobHistoryIndex every time recordJobHistory is called.
If so, You can replace everything inside try block with following compact approach in this method and make it argument-less.
```
CreateIndexRequest cir = new CreateIndexRequest(JOB_HISTORY_INDEX_NAME).mapping(
                    historyMapping(),
                    (MediaType) XContentType.JSON
                ).settings(INDEX_SETTINGS).waitForActiveShards(1);
                ActionListener<CreateIndexResponse> cirListener = ActionListener.wrap(response -> {
                    ctx.restore();
                    LOGGER.info("Job History index {} created.", JOB_HISTORY_INDEX_NAME);
                }, (failResponse) -> {
                    /* Index already exists, ignore and continue */
                    LOGGER.info("Index {} already exists.", JOB_HISTORY_INDEX_NAME);
                });
                this.client.admin().indices().create(cir, cirListener);
```

spi/src/main/java/org/opensearch/jobscheduler/spi/utils/JobHistoryService.java
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    private void createHistoryRecord(final StatusHistoryModel historyModel, ActionListener<Boolean> listener) {
+        try (ThreadContext.StoredContext ignore = client.threadPool().getThreadContext().stashContext()) {
+            String historyId = generateHistoryId(historyModel.getJobIndexName(), historyModel.getJobId(), historyModel.getStartTime());
+
+            final IndexRequest request = new IndexRequest(JOB_HISTORY_INDEX_NAME).id(historyId)
+                .source(historyModel.toXContent(XContentFactory.jsonBuilder(), ToXContent.EMPTY_PARAMS))
+                .setIfSeqNo(SequenceNumbers.UNASSIGNED_SEQ_NO)
+                .setIfPrimaryTerm(SequenceNumbers.UNASSIGNED_PRIMARY_TERM)
```
DarshitChanpura, line 162: why are we deliberately assigning unassigned primary term?

spi/src/main/java/org/opensearch/jobscheduler/spi/utils/JobHistoryService.java
```diff
@@ -0,0 +1,263 @@
...
+                    }
+                }, listener::onFailure));
+            } else {
+                listener.onResponse(false);
+            }
+        }, listener::onFailure));
+    }
+
+    private void createHistoryRecord(final StatusHistoryModel historyModel, ActionListener<Boolean> listener) {
+        try (ThreadContext.StoredContext ignore = client.threadPool().getThreadContext().stashContext()) {
```
DarshitChanpura, line 156: we can skip stashing context here since it is only called inside findHistoryRecord which stashes context.

spi/src/main/java/org/opensearch/jobscheduler/spi/utils/LockService.java
```diff
@@ -125,7 +133,25 @@ public void acquireLock(
...
+                        null,
+                        1,
+                        ActionListener.wrap(success -> listener.onResponse(lock), failure -> listener.onResponse(lock))
+                    );
+                } else {
+                    listener.onResponse(lock);
+                }
+            } else {
+                listener.onResponse(null);
+            }
```
DarshitChanpura, line 153: can be simplified to:
```
if (lock != null && historyService != null) {
                    historyService.recordJobHistory(
                        jobIndexName,
                        jobId,
                        lock.getLockTime(),
                        null,
                        1,
                        ActionListener.wrap(success -> listener.onResponse(lock), failure -> listener.onResponse(lock))
                    );
                } else {
                    listener.onResponse(lock);
                }

**Review by DarshitChanpura — COMMENTED** [review]
Thank you @Jeremydupras . PR looks close. left few comments. Will take a second pass once all are addressed.

File: spi/src/test/java/org/opensearch/jobscheduler/spi/utils/JobHistoryServiceIT.java [hunk 8]
```diff
@@ -0,0 +1,255 @@
...
+            assertTrue("Failed to record initial job history", result);
+
+            // Update with end time
+            historyService.recordJobHistory(jobIndexName, jobId, startTime, endTime, 2, ActionListener.wrap(updateResult -> {
+                assertTrue("Failed to update job history", updateResult);
+                latch.countDown();
+            }, exception -> fail("Exception during update: " + exception.getMessage())));
+        }, exception -> fail("Exception during initial record: " + exception.getMessage())));
+
+        latch.await(15L, TimeUnit.SECONDS);
```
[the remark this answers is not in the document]
DarshitChanpura: we should not await tests. You handle assertions in the listener block so that should cover the case.

spi/src/main/java/org/opensearch/jobscheduler/spi/utils/LockService.java
```diff
@@ -45,21 +45,38 @@
...
+        this.clusterService = clusterService;
+        this.historyService = historyService;
+        this.statusHistoryEnabled = statusHistoryEnabled;
+    }
+
     public LockService(final Client client, final ClusterService clusterService) {
         this.client = client;
         this.clusterService = clusterService;
+        this.historyService = null;
+        this.statusHistoryEnabled = null;
```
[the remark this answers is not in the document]
Jeremydupras: changed

.codecov.yml
```diff
@@ -4,11 +4,11 @@ codecov:
 coverage:
   precision: 2
   round: down
-  range: "75...100"
+  range: "70...100"
   status:
     project:
       default:
-        target: 75%    # the required coverage value
+        target: 70%    # the required coverage value
```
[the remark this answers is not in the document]
cwperks: 70% is what security repo is set at. We can set the bar higher as the coverage in the repo increases. Before the summer coverage was errantly showing 30% because this repo wasn't equipped with dumping coverage from testclusters. It fixed now and the repo is nearly at 70%.

spi/src/main/java/org/opensearch/jobscheduler/spi/JobExecutionContext.java
```diff
@@ -43,6 +45,7 @@ public JobExecutionContext(StreamInput in) throws IOException {
         this.lockService = null;
         this.jobIndexName = in.readString();
         this.jobId = in.readString();
+        this.jobStatus = in.readInt();
```
cwperks, line 48: Changing serialization can create bwc problems.

We need to surround this and similar line in writeTo with version guards

```
if (in.getVersion().onOrAfter(Version.3_3_0)) {
    this.jobStatus = in.readInt();
}
```

spi/src/main/java/org/opensearch/jobscheduler/spi/StatusHistoryModel.java
```diff
@@ -0,0 +1,186 @@
...
+    public static final String START_TIME = "start_time";
+    public static final String END_TIME = "end_time";
+    public static final String COMPLETION_STATUS = "completion_status";
+
+    private final String jobIndexName;
+    private final String jobId;
+    private final Instant startTime;
+    private final Instant endTime;
+    private final int status;
+    private final long seqNo;
```
cwperks, line 38: Do we need to keep track of `seqNo` and `primaryTerm`. I'd prefer to only keep references to the new logic being introduced and not keep track of these metadata fields.

spi/src/main/java/org/opensearch/jobscheduler/spi/utils/JobHistoryService.java
```diff
@@ -0,0 +1,261 @@
...
+        return clusterService.state().routingTable().hasIndex(JOB_HISTORY_INDEX_NAME);
+    }
+
+    void createHistoryIndex(ActionListener<Boolean> listener) {
+        try (ThreadContext.StoredContext ignore = client.threadPool().getThreadContext().stashContext()) {
+            if (historyIndexExist()) {
+                listener.onResponse(true);
+            } else {
+                final CreateIndexRequest request = new CreateIndexRequest(JOB_HISTORY_INDEX_NAME).mapping(
+                    historyMapping(),
```
cwperks, line 79: I think its fine to declare the mappings, but I'm dealing with a problem in ISM that also creates mappings on index creation and then does not allow new mappings to be added in subsequent minor versions. 

Let's keep this here, but if we were to add a new field (say `exception_message`) then how would that work on upgrade?

spi/src/main/java/org/opensearch/jobscheduler/spi/utils/LockService.java
```diff
@@ -45,21 +45,38 @@
...
+import java.util.function.Supplier;
 
 public class LockService {
     private static final Logger logger = LogManager.getLogger(LockService.class);
     public static final String LOCK_INDEX_NAME = ".opendistro-job-scheduler-lock";
 
     private final Client client;
     private final ClusterService clusterService;
-    final static Map<String, Object> INDEX_SETTINGS = Map.of("index.number_of_shards", 1, "index.auto_expand_replicas", "0-1");
+    final static Map<String, Object> INDEX_SETTINGS = Map.of("index.number_of_shards", 1, "index.auto_expand_replicas", "0-all");
```
cwperks, line 56:
```suggestion
    final static Map<String, Object> INDEX_SETTINGS = Map.of("index.number_of_shards", 1, "index.auto_expand_replicas", "0-1");
```

src/main/java/org/opensearch/jobscheduler/JobSchedulerSettings.java
```diff
@@ -53,4 +53,11 @@ public class JobSchedulerSettings {
         Setting.Property.NodeScope,
         Setting.Property.Dynamic
     );
+
+    public static final Setting<Boolean> STATUS_HISTORY = Setting.boolSetting(
+        "plugins.jobscheduler.status_history",
```
cwperks, line 58:
```suggestion
        "plugins.jobscheduler.history.enabled",
```

src/main/java/org/opensearch/jobscheduler/JobSchedulerSettings.java
```diff
@@ -53,4 +53,11 @@ public class JobSchedulerSettings {
...
+        false,
+        Setting.Property.NodeScope,
```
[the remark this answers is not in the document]
Jeremydupras: Removing the Nodescope causes a compile time error.
java.lang.IllegalArgumentException: No scope found for setting [plugins.jobscheduler.status_history]
opensearch-node2       | uncaught exception in thread [main]

src/main/java/org/opensearch/jobscheduler/JobSchedulerSettings.java
Jeremydupras: changed

spi/src/main/java/org/opensearch/jobscheduler/spi/utils/LockService.java
Jeremydupras: changed

spi/src/main/java/org/opensearch/jobscheduler/spi/utils/JobHistoryService.java
```diff
@@ -0,0 +1,261 @@
...
+import java.time.Instant;
+import java.util.Map;
+
+public class JobHistoryService {
+    private static final Logger logger = LogManager.getLogger(JobHistoryService.class);
+    public static final String JOB_HISTORY_INDEX_NAME = ".opendistro-job-scheduler-history";
+
+    private final Client client;
+    private final ClusterService clusterService;
+    final static Map<String, Object> INDEX_SETTINGS = Map.of("index.number_of_shards", 1, "index.auto_expand_replicas", "0-all");
```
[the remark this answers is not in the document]
Jeremydupras: changed

spi/src/main/java/org/opensearch/jobscheduler/spi/utils/JobHistoryService.java
```diff
@@ -0,0 +1,261 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import java.time.Instant;
+import java.util.Map;
+
+public class JobHistoryService {
+    private static final Logger logger = LogManager.getLogger(JobHistoryService.class);
+    public static final String JOB_HISTORY_INDEX_NAME = ".opendistro-job-scheduler-history";
```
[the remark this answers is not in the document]
Jeremydupras: changed

spi/src/test/java/org/opensearch/jobscheduler/spi/utils/JobHistoryServiceIT.java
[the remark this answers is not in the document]
Jeremydupras: The historyService has asynchronous methods when writing to in Index. The Await ensures that the method finished before evaluating the result.

spi/src/main/java/org/opensearch/jobscheduler/spi/JobExecutionContext.java
Jeremydupras: removed from job execution context

spi/src/main/java/org/opensearch/jobscheduler/spi/JobExecutionContext.java
```diff
@@ -51,6 +54,7 @@ public void writeTo(StreamOutput out) throws IOException {
         this.jobVersion.writeTo(out);
         out.writeString(this.jobIndexName);
         out.writeString(this.jobId);
+        out.writeInt(this.jobStatus);
```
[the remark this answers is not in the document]
Jeremydupras: removed from job execution context

spi/src/main/java/org/opensearch/jobscheduler/spi/StatusHistoryModel.java
Jeremydupras: The current design adds to the history list when the lock is created and updates the entry when the lock in released. If the sequence number and primary term are not tracked the history service should be changed to append only.

spi/src/main/java/org/opensearch/jobscheduler/spi/JobExecutionContext.java
```diff
@@ -22,6 +22,7 @@ public class JobExecutionContext implements Writeable {
     private final LockService lockService;
     private final String jobIndexName;
     private final String jobId;
+    private Integer jobStatus;
```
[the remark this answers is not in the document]
Jeremydupras: Nope, it has been removed

spi/src/main/java/org/opensearch/jobscheduler/spi/utils/JobHistoryService.java
Jeremydupras: If the History service is toggled by the Enabled History setting. Wouldn't it be better to create the history index in the history service while this is still optional service? This would ensure the index is created if the history service was enabled during runtime. If the service becomes a permanent addition then moving the index creation to node start would make sense.

cwperks (approved):

